PR #16, opened 2024-03-03
Readme file updated

KonstantinVchkov:
Added installation instructions in the readme file.


edichoska (approved):
we should update the README in the future. But it is a good start

PR #17, opened 2024-03-04
ESLint, babel & husky init setup

MarkoDavkovski:


KonstantinVchkov (commented):
I've reviewed the ESLint, Babel, and Husky setup, and I'm pleased to say that everything is excellently configured and aligns perfectly with our project's standards and requirements.

.husky/pre-commit
```diff
@@ -0,0 +1 @@
+npm run lint
```
edichoska, line 1: we should be checking only the staged files

- npm lint-staged
- you would need to install lint-staged as well

Also lets take a note that after adding prettier we should include it here as well

edichoska (approved):
Looks great. You can merge it 👏

PR #35, opened 2024-03-05
Migrated to app router structure

kire21:
Created: 
- App folder(layout, page, global.scss and home.module.scss files)

Deleted:
- src folder(all files are inside)
- babelrc file (Resolving "Babel and `next/font` Conflict", source: https://nextjs.org/docs/messages/babel-font-loader-conflict)


edichoska (approved):
🥇

PR #56, opened 2024-03-08
Implemented global loading component

ninman27:
I created a global loading file.
The component is implemented in the global layout.

### DIFF
--- a/app/loading.tsx
+++ b/app/loading.tsx
@@ -1,0 +1,5 @@
+const Loading = () => {
+  return <div>Loading...</div>;
+};
+
+export default Loading;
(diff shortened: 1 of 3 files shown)


app/loading.tsx
```diff
@@ -0,0 +1,5 @@
+const Loading = () => {
+  return <div>Loading...</div>;
+};
+
+export default Loading;
```
edichoska, line 5: Can you create a spinner that would be fixed, it will be centered vertically and horizontally.
So we would have the spinner, and <p> Loading... </p>under it.
The spinner should have a default color, but we should be able to change the color.
For the look I am thinking something like this, but if you find a better looking option feel free to implement it.
1_6Pv-vKJnzLZNn1iR1GmFtg

edichoska (approved):
Looks good 👏

PR #58, opened 2024-03-08
setup mui

kire21:
- Installed MUI Library
- Created Button Component (changing style through props)
- Created Card Component


edichoska (commented):
As we discussed yesterday, we wont be using MUI for this project.

PR #61, opened 2024-03-10
Configuration project setup for Prettier,ESLint & Husky

MarkoDavkovski:


app/loading.tsx
```diff
@@ -1,12 +1,12 @@
-import style from "./loading.module.scss";
+import style from './loading.module.scss';
 
-const Loading = () => {
+function Loading() {
```
RevanchistX, line 3: Why is this done?
This change seems uneccessary.

README.md
```diff
@@ -4,18 +4,17 @@ LearnHub.mk is an innovative online learning platform dedicated to nurturing the
 
 Our mission is to provide a collaborative space where individuals can enhance their web development expertise, work on real projects, and engage with a supportive community. Whether you are embarking on your coding journey or a seasoned developer, LearnHub.mk has something valuable to offer.
 
-
 ## Features
 
 ### 📚 Comprehensive Learning Resources
 
-Access a diverse range of educational materials covering various web development topics. 
+Access a diverse range of educational materials covering various web development topics.
```
RevanchistX, line 11: Updating the Readme is fine, but the changes in this Pull Request have nothing to do with Prettier.
Leave it like this, but for future reference -  make smaller commits based on functionality.

MarkoDavkovski: https://stackoverflow.com/questions/75587906/why-airbnb-style-guide-force-eslint-rule-react-function-component-definition

app/loading.tsx
MarkoDavkovski: https://stackoverflow.com/questions/75587906/why-airbnb-style-guide-force-eslint-rule-react-function-component-definition

edichoska: Davko lets chat on the next meeting about implementing a style guide. THis tiket is just for prettier and adding it into husky

edichoska (commented):
Prettier config, and husky looks good. Just please revert the changes to the README, and the loading componenet, and rerequest a review.

app/loading.tsx
RevanchistX: @MarkoDavkovski As @edichoska said, it's important to be consistent throughout the project, regardless of which approach is picked.

If this change is done by eslint, we should investigate the rule that forces this change and maybe disable it.

MarkoDavkovski: Absolutely, consistency across the project is crucial, as both @edichoska and the ticket description emphasize. We opted for the Airbnb style guide, which indeed mandates the use of function components over arrow functions for components.
If there are areas where my implementation can be enhanced or if the application of the Airbnb rules through eslint has unintentionally enforced any undesirable changes, I'm open to suggestions.

RevanchistX: I wasn't aware of the Airbnb style guide decision.
I have nothing against your changes other than the fact that I didn't understand them and wanted to make sure everything is according to plan. 

You're on the right track. Please proceed with this so we can finalize this change.
https://github.com/learnhubmkd/app/pull/61#pullrequestreview-1926549314

edichoska (approved):

PR #62, opened 2024-03-10
Update README.md

Mikivitkovski:
I update the Readme.md file because I make a new file, and I commit to the project, and that file needs description to work properly, to install a specific version of some package


README.md
```diff
@@ -64,7 +64,13 @@ Before diving into LearnHub.mk, make sure you have Node.js installed on your mac
 
        yarn dev
 
-4. Open [localhost:3000](http://localhost:3000) in your browser to explore the app.
+4. Install specific version of Node.js
+
+           nvm install $(cat .nvmrc)
```
RevanchistX, line 69: This should be the the first thing done in the project, not the last.

RevanchistX: Additionally, please include instructions for Windows Linux and Mac

Mickeyy23: Other things what we need to install with Node are already included before

.nvmrc
```diff
@@ -0,0 +1 @@
+21.6.2
```
edichoska, line 1: We should add the Long-Term Support (LTS) version of Node.js rather than the current/latest version.
You should also update the package.json and include the "engine"

README.md
edichoska: You should add the Node version in a 
## Prerequisites

- Node.js version:


and also add they should run `nvm use` before npm install

Mikivitkovski: I make some changes in Readme.md file and like i research we dont need explicitly run `nvm use` after installing the specific Node.js version using `nvm install $(cat .nvmrc)`. The `nvm use` command is automatically handled by the `.nvmrc` file when you navigate to the project.

edichoska (approved):

PR #66, opened 2024-03-11
Jest

DimitarHristovski:


components/test/component.test.js
```diff
@@ -0,0 +1,3 @@
+test('Jest Test Example', () => {
+  expect(true).toBe(true);
+});
```
edichoska, line 3: we are using Typescript in the project. Please use the .ts instead of .js

stoimenov95: @DimitarHristovski can you please provide another example related to the app itself, e.g. a test for some of the existing components (Render the component with some text prop, assert that the expected text is in the document)

stoimenov95 (changes requested):

components/test/component.test.js
DimitarHristovski: Theres no code to be tested yet or if there is give me an example and im gonna change it

stoimenov95: @DimitarHristovski you can add a test for the page component in the app root folder and check if the h1 is rendering with the text 'Home LearnHub.mk'. Everything else regarding the setup seems all right to me, @edichoska what do you say?

PR #79, opened 2024-03-11
setup routing

ninman27:
Created route groups:

- website (index, blog, single blog post)
- admin (admin-dashboard)
- auth (login, signup)
- platform (member-dashboard)


edichoska (commented):
The only concern i am having with route groups is:

> With Routes that include a route group should not resolve to the same URL path as other routes. For example, since route groups don't affect URL structure, (marketing)/about/page.js and (shop)/about/page.js would both resolve to /about and cause an error.

Lets use them only when we would need a specific layout for a route.

edichoska (approved):
Looks great now! You can merge it 👏

PR #85, opened 2024-03-13
fixed error (function component is not a function declaration)

ninman27:
I made changes in all routes for function type.


dichoska:
Since we wont be using airbnb style guide and declaring components as functions, this wouldn't be an issue.
Fix for this comming shortly. @ninman27

PR #86, opened 2024-03-13
Implemented global error and 404 page

ninman27:
global-errors.tsx working only in production, not in development...


edichoska (approved):
🍀

PR #87, opened 2024-03-13
Ticket 68 and 69 create footer and package.json and Navigation from y…

Mikivitkovski:
…esterday updated


app/layout.tsx
```diff
@@ -14,15 +17,16 @@ export const metadata: Metadata = {
...
   children: React.ReactNode;
 }>) {
   return (
     <html lang="en">
       <body className={roboto.className}>
         <main>
+          <Navigation />
           <Suspense fallback={<Loading />}>{children}</Suspense>
         </main>
+        <Footer />
```
edichoska, line 29: The navigation should be placed outside of main.
Read more about main tag here: https://www.w3schools.com/tags/tag_main.asp

components/reusable-components/footer.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <div className={styles.contactSection}>
+        <p>CONTACT</p>
+        <p>contact@learnhub.mk</p>
+      </div>
+
+      <div className={styles.socialMediaSection}>
+        <p>SOCIAL MEDIA</p>
+        <Link href="https://www.linkedin.com/" target="_blank" rel="noopener noreferrer">
+          <FaLinkedin /> LINKEDIN
+        </Link>
```
edichoska, line 22: All Social Media Links are uppercase. You can use text-transform: uppercase to achieve that.

components/reusable-components/navigation.tsx
```diff
@@ -0,0 +1,43 @@
...
+                <Link href="/"> Цели </Link>
+              </li>
+              <li className={styles.navLink1}>
+                <Link href="/"> Контакт </Link>
+              </li>
+              <li className={styles.navLink1}>
+                <Link href="/">Блог </Link>
+              </li>
+            </ul>
+          </div>
```
edichoska, line 28: we do not need a list here.
Also this div should be centered in the page.

components/reusable-components/navigation.tsx
```diff
@@ -0,0 +1,43 @@
...
+        </div>
+        <div className={styles.rightContent}>
+          <button className={styles.customButton1} type="submit">
+            Sign In
+          </button>
+          <button className={styles.customButton} type="submit">
+            Sign Up
+          </button>
```
edichoska, line 36: Create reusable Button component.
It should have default styles, but we should be able to override those styles.
It should receive props such as background color, color, text, border, border radius

edichoska (commented):
@Mikivitkovski  
Each component should be in its own folder.
Create a folder Footer and add footer.tsx and footer.module.scss. Same for navigation.
Also follow the naming convention i mentioned yesterday:  PascalCase

Issue #68, opened 2024-03-11
[FE] Create the Navigation


edichoska:
**Task:**
- Implement the Navigation as shown in the Figma file:  https://www.figma.com/file/zhRSxKmdeXlsdStdL6NdgU/Untitled?type=design&node-id=0-1&mode=design&t=VgX6o5TWVv38P8ns-0
- I will attach the logo file as soon as we have it, for now you can just type LearnHub

PR #88, opened 2024-03-13
#68 create navigation

Mikivitkovski:


edichoska:
If you have a conflict on a branch you should resolve it on the same branch.
For example here you should have merged main branch into your branch, and resolved the conflict.

Issue #75, opened 2024-03-11
[FE] Create the Blogs Section


edichoska:
**Task:**
- Implement the Blogs Section as shown in the Figma file: https://www.figma.com/file/zhRSxKmdeXlsdStdL6NdgU/Untitled?type=design&node-id=0-1&mode=design&t=VgX6o5TWVv38P8ns-0
- When we click on the button we should be redirected to the Blogs Page
- When we click on the card we should be redirected to the Individual Blog Page
(at this phase do not worry about responsiveness)

kire21:
![Image](https://github.com/learnhubmkd/app/assets/58231146/a03a61b8-ba83-43c0-85b4-c48bd2ecdfcf)



![Image](https://github.com/learnhubmkd/app/assets/58231146/94017135-d4bb-4961-a5f5-3f42c47a5f40)

PR #92, opened 2024-03-14
Blog section #75

kire21:


components/reusable-components/BlogCard/BlogCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+function BlogCard({ id, title, body }: BlogCardProps) {
+  const truncate = ({ str, noWords }: TruncateProps) => {
+    return str.split(' ').splice(0, noWords).join(' ');
+  };
+  return (
+    <div key={id} className={style.cardContainer}>
+      <div>
+        <h2 className={`headline-m ${style.cardTitle}`}>{title}</h2>
```
edichoska, line 24: can you shorten the title, so it is always in two lines?

components/reusable-components/BlogCard/BlogCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+interface TruncateProps {
+  str: string;
+  noWords: number;
+}
+
+function BlogCard({ id, title, body }: BlogCardProps) {
+  const truncate = ({ str, noWords }: TruncateProps) => {
+    return str.split(' ').splice(0, noWords).join(' ');
```
edichoska, line 19: Although the use of splice here won't cause any side effects, it is better to use slice since it does not modify the original array but instead returns a new array. (pure functions)

components/module-components/BlogList/BlogList.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+async function BlogList() {
+  const data = await getPosts();
+  return (
+    <>
+      <div className={`grid grid__1x3 ${style.containerMargin}`}>
+        {data.posts?.slice(0, 6).map((post: { id: string; title: string; body: string }) => {
+          return <BlogCard id={post?.id} title={post?.title} body={post?.body} />;
+        })}
+      </div>
```
edichoska, line 21: If we do not get any data, or there is an error can we display 'Нема блог постови во моментов' (or some better wording in this context).

edichoska (commented):
@kire21  Can you please provide a screenshot when you are done with the changes?

edichoska (approved):
🚀

PR #93, opened 2024-03-14
Set ESLint to use arrow function components

MarkoDavkovski:


edichoska (approved):
thank you for fixing this 🙏

Issue #69, opened 2024-03-11
[FE] Create the Footer


edichoska:
**Task:**
- Implement the Footer as shown in the Figma file: https://www.figma.com/file/zhRSxKmdeXlsdStdL6NdgU/Untitled?type=design&node-id=0-1&mode=design&t=VgX6o5TWVv38P8ns-0
(right now do not worry about responsiveness)

PR #95, opened 2024-03-14
#69 Problem bug solved and everything work

Mikivitkovski:


app/styles/globals/_colors.scss
```diff
@@ -1,5 +1,5 @@
 :root {
-  --bg-color: hsl(0, 0%, 11%);
+  --bg-color: #202023;
   --text-color: hsl(0, 0%, 100%);
```
edichoska, line 3: why is this changed?

Mikivitkovski: In figma is that code color for background and nav and footer i make my nav and footer in color what is in figma and i see difference in colors and i check again and i see that are not same colors in figma and in _colors.scss

components/reusable-components/navigation/navigation.module.scss
```diff
@@ -0,0 +1,64 @@
...
+
+.navLink1 {
+  color: #ffff;
+  font-size: 20px;
+  margin-right: -400px;
+  margin-left: 450px;
+
+  /* Adjust the margin for space between links */
+}
+
```
edichoska, line 23: remove all the comments in this file

Issue #73, opened 2024-03-11
[FE] Create the Newsletter Section


edichoska:
**Task:**
- Implement the Newsletter Section as shown in the Figma file: https://www.figma.com/file/zhRSxKmdeXlsdStdL6NdgU/Untitled?type=design&node-id=0-1&mode=design&t=VgX6o5TWVv38P8ns-0
- Add validation to the input, make sure we are sending an email
- We should also show a notification on the screen to show that the action was successful or if it failed
(at this phase do not worry about responsiveness)

PR #101, opened 2024-03-19
#73 Create Newsletter section and update newsletter section in footer

Mikivitkovski:
pr
image


app/layout.tsx
```diff
@@ -20,6 +21,7 @@ const RootLayout = ({
       <main>
         <Suspense fallback={<Loading />}>{children}</Suspense>
       </main>
+      <Newsletter />
```
edichoska, line 24: The Newsletter shouldn't be part of the layout. It is not a component that should appear on all pages.
Here we have the Navigation and the Footer, because they will appear on all pages.

edichoska (commented):
Just make the small changes, and please include a screenshot

app/layout.tsx
Mikivitkovski: Okay i change,  i put in page.tsx file under BlogList

components/reusable-components/newsletter/emailform/emailform.tsx
```diff
@@ -0,0 +1,66 @@
...
+      setShowNotification(true);
+    }
+  };
+
+  return (
+    <div>
+      {showNotification && (
+        <Notification
+          message={notificationMessage}
+          type={notificationType}
```
edichoska, line 36: In this case this is not a bad approach since we only have one input. 
It is better to have the formik initial values, validation schema outside.
We would have const formik = useFormik<InitialValues>({
 <here you would have onSubmit function where you would pass the values>
 and in the component you can do:
`     <form onSubmit={formik.handleSubmit}>`

edichoska:
Ah also I noticed that the button with the arrow is a pixel or two bigger than the input @Mikivitkovski

stoimenov95:
Screenshot 2024-03-21 at 20 39 35
@Mikivitkovski do I notice some bigger space in the footer between the Learnhub.mk and the Contact section?

stoimenov95:
@Mikivitkovski 
Screenshot 2024-03-21 at 20 42 31
Please validate the email field to accept only valid e-mail addresses.

edichoska:
> Screenshot 2024-03-21 at 20 39 35 @Mikivitkovski do I notice some bigger space in the footer between the Learnhub.mk and the Contact section?

we are not doing responsive at this time.

components/reusable-components/newsletter/emailform/emailform.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const EmailForm: React.FC<EmailFormProps> = ({ inputClassName, buttonClassName }) => {
+  const [showNotification, setShowNotification] = useState(false);
+  const [notificationMessage, setNotificationMessage] = useState('');
+  const [notificationType, setNotificationType] = useState<'success' | 'error'>('success');
+
+  const formik = useFormik({
+    initialValues: { email: '' },
+    validationSchema: Yup.object({
+      email: Yup.string().email('Invalid email address!').required('Email is required!'),
```
edichoska, line 22: you will need a custom regex here since this validation won't be enough.
and then do .matches with that regex

edichoska:
> > > Screenshot 2024-03-21 at 20 39 35 @Mikivitkovski do I notice some bigger space in the footer between the Learnhub.mk and the Contact section?
> > 
> > 
> > we are not doing responsive at this time.
> 
> it's not regarding the responsiveness, it's on the default resolution.

yes you are right. its somehow broken.  @Mikivitkovski  also the navigation is broken as well

Mikivitkovski:
> Screenshot 2024-03-21 at 20 39 35 @Mikivitkovski do I notice some bigger space in the footer between the Learnhub.mk and the Contact section?

Its already validated check the update of the file not first version

stoimenov95:
@Mikivitkovski you copied the emailform content in the same file and the project won't start. Please fix it and carefully look through your changes by comparing the previous and next versions before committing and pushing to the remote.

components/reusable-components/newsletter/emailform/emailform.tsx
```diff
@@ -16,16 +16,19 @@ const EmailForm: React.FC<EmailFormProps> = ({ inputClassName, buttonClassName }
   const [notificationMessage, setNotificationMessage] = useState('');
   const [notificationType, setNotificationType] = useState<'success' | 'error'>('success');
 
+  // Regular expression for basic email validation
```
edichoska, line 19: remove comment

edichoska (changes requested):
notification should be its own component and not a part of emailform

edichoska (approved):
🚀

PR #102, opened 2024-03-19
#102 add reposnive fonts

GregoriBavaro:
Add responsive CSS fonts variables within the :root in  app>styles>globals>_typography.scss

These custom properties are used to define font sizes at different breakpoints using the clamp() function along with vw (viewport width) units.

Each clamp() function call ensures that the font sizes adapt smoothly between the defined minimum and maximum sizes based on the viewport width.


edichoska:
Let's leave this one for when we get the fonts and all the font sizes used by the UI/UX team.

RevanchistX:
@GregoriBavaro Please include ticket number in the pull request name

Issue #74, opened 2024-03-11
[FE] Create the Member Spotlight Section


edichoska:
**Task:**
- Implement the Member Spotlight Section as shown in the Figma file: https://www.figma.com/file/zhRSxKmdeXlsdStdL6NdgU/Untitled?type=design&node-id=0-1&mode=design&t=VgX6o5TWVv38P8ns-0
- We should be able to scroll and drag thought the cards (explore solutions for this)
(at this phase do not worry about responsiveness)

PR #130, opened 2024-03-28
#74 created the member spotlight section

ninman27:
Installed Embla carousel library
Created member cards
Created a carousel with Embla
prs


components/module-components/member-list/MemberList.tsx
```diff
@@ -0,0 +1,72 @@
...
+      memberSkills: ['NextJS', 'SCSS', 'TypeScript', 'PR1', 'PR2', 'PR3'],
+      memberProjects: ['AxM', '', ''],
+    },
+    {
+      memberName: 'Pero',
+      memberInfo:
+        'Lorem ipsum dolor, sit amet consectetur adipisicing elit. Incidunt quas repudianda delectus dolorum tempora.',
+      memberSkills: ['PhP', 'NodeJS', 'JavaScript', 'PR1', 'PR3', 'PR4'],
+      memberProjects: ['Pero.mk', '', ''],
+    },
```
edichoska, line 29: please extract this data outside of the component

components/module-components/member-list/MemberList.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <div className={style.carouselContainer} ref={emblaRef}>
+        <div className={style.carouselCardsContainer}>
+          {FakeMemberData.members.map((member) => (
+            <MemberCard
+              key={member.memberName}
+              memberName={member.memberName}
+              memberInfo={member.memberInfo}
+              memberSkills={member.memberSkills}
+              memberProjects={member.memberProjects}
+            />
```
edichoska, line 64: We would also need an image here. So the user when signing up could optionally upload an image. We would check if the user has an image, if not we would show an avatar.
For now you can include memberImage="' in the fake data, and if it is empty or null, we would show the image you have included, as a backup

edichoska:
@ninman27 the code looks great!
Some notes about the design:
1. Could you play a little bit with the background color of the cards, so there is bigger contract with the background of the page
2. Could you maybe include dots under the cards so the user knows she/he can scroll

edichoska:
@stoimenov95 i am not sure if we should make the slider scrolling infinite. meaning when the user comes to the last card, the first one appears. But it looks good this way as well

stoimenov95 (approved):

ninman27:
MemberSpotlightSection

stoimenov95:
> @stoimenov95 i am not sure if we should make the slider scrolling infinite. meaning when the user comes to the last card, the first one appears. But it looks good this way as well

yep, that is a great idea.

components/reusable-components/embla-carousel-dot-button/EmblaCarouselDotButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+    scrollSnaps,
+    onDotButtonClick,
+  };
+};
+
+export const DotButton = (props: any) => {
+  const { children, ...restProps } = props;
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    <div type="button" {...restProps}>
```
stoimenov95, line 52: @ninman27 The DotButton is using a div with a type="button" attribute, but div elements don't have a type attribute. For accessibility, consider using a button element instead. This will make it focusable and interactive for users relying on keyboard navigation without extra effort.

components/reusable-components/embla-carousel-dot-button/EmblaCarouselDotButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const onSelect = useCallback((emblaApiDots: EmblaCarouselType) => {
+    setSelectedIndex(emblaApiDots.selectedScrollSnap());
+  }, []);
+
+  useEffect(() => {
+    if (!emblaApi) return;
+
+    onInit(emblaApi);
+    onSelect(emblaApi);
+    emblaApi.on('reInit', onInit);
```
stoimenov95, line 36: @ninman27 we can use embaApi.off() in the return function of the `useEffect` in order to do a cleanup?

stoimenov95 (changes requested):
@ninman27 The functionality works great on my side. Try to make the carousel to show the first item after we get to the end, to keep the user interacting.

components/reusable-components/embla-carousel-dot-button/EmblaCarouselDotButton.tsx
ninman27: Done

stoimenov95:
@ninman27 did you have some time to look into an infinite scroll option for the carousel?

components/reusable-components/embla-carousel-dot-button/EmblaCarouselDotButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+    emblaApi.on('reInit', onSelect);
+    emblaApi.on('select', onSelect);
+
+    // eslint-disable-next-line consistent-return
+    return () => {
+      emblaApi.off('reInit', onInit);
+      emblaApi.off('reInit', onSelect);
```
stoimenov95, line 43: @ninman27 don't we need one more `emblaApi.off('select', onSelect);` ?

stoimenov95 (approved):
Perfect @ninman27 🚀

Issue #76, opened 2024-03-11
[FE] Create the Contact Section


edichoska:
**Task:**
- Implement the Contact Section as shown in the Figma file: https://www.figma.com/file/zhRSxKmdeXlsdStdL6NdgU/Untitled?type=design&node-id=0-1&mode=design&t=VgX6o5TWVv38P8ns-0
- Add validation for the inputs
- We should show error messages as soon as the field is defocused(research this)
- We should also show a notification on the screen to show that the action was successful or if it failed
(at this phase do not worry about responsiveness)

PR #136, opened 2024-04-04
#76 contact section

aleksandarmicev:
Implement the Contact section, The Form is made with showing alert if some input in unfiled,
Also the is alert for success submitting the form.
In this section, still persist problem with the logo "Let's collaborate", therefore there is no code for that part.


app/page.module.css
```diff
@@ -0,0 +1,212 @@
...
+  display: inherit;
+  justify-content: inherit;
+  align-items: inherit;
+  font-size: 0.85rem;
+  max-width: var(--max-width);
+  width: 100%;
+  z-index: 2;
+  font-family: var(--font-mono);
+}
+
```
edichoska, line 20: Can you please use more descriptive names of your classes?

app/page.tsx
```diff
@@ -1,12 +1,14 @@
...
-import BlogList from '../components/module-components/blog-list/BlogList';
+import Contact from '../components/module-components/Contact/Contact';
 import LearnhubAnimation from '../components/module-components/learnhub-animation/LearnhubAnimation';
-import styles from './page.module.scss';
 import Newsletter from '../components/reusable-components/newsletter/Newsletter';
 import MemberList from '../components/module-components/member-list/MemberList';
-
 import FAQ_DUMMY_DATA from '../data/FAQ';
+import styles from './page.module.scss';
+// import BlogList from '../components/module-components/blog-list/BlogList';
```
edichoska, line 11: Why is the BlogList commented?

components/module-components/Contact/Contact.tsx
```diff
@@ -0,0 +1,141 @@
...
+        <div className={style.collabInside}>
+          <svg
+            viewBox="0 0 64 64"
+            width="30"
+            height="30"
+            xmlns="http://www.w3.org/2000/svg"
+            xmlnsXlink="http://www.w3.org/1999/xlink"
+            aria-hidden="true"
+            role="img"
+            // className="iconify iconify--emojione-monotone"
```
edichoska, line 59: remove comment

components/module-components/Contact/Contact.tsx
```diff
@@ -0,0 +1,141 @@
...
+                  attributeType="xml"
+                  attributeName="transform"
+                  type="rotate"
+                  from="0"
+                  to="360"
+                  dur="4.5s"
+                  additive="sum"
+                  repeatCount="indefinite"
+                />
+                <defs>
```
edichoska, line 33: You can extract the Circle in its own component

components/module-components/Contact/Contact.tsx
```diff
@@ -0,0 +1,141 @@
...
+            if (!values.email) {
+              errors.email = 'Задолжително внесете емаил адреса';
+            } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/i.test(values.email)) {
+              errors.email = 'Невалидна емаил адреса';
+            }
+            if (!values.message) {
+              errors.message = 'Задолжително внесете порака';
+            }
+            return errors;
+          }}
```
edichoska, line 97: use yup for validation, and create a validation schema. 
you should make sure the user name is a string etc.
check the newsletter component it has something similar in it.
Also the error messages should look the same as the ones in newsletter

edichoska (changes requested):

edichoska:
@aleksandarmicev  The animation and the sticker are looking great! Good job on that. 👏 

Some additional notes:
1. Can you make sure the font size is larger so it occupies the whole circle, as shown in the figma file.(of the text in the circle animation)
2. Also, the form shouldn't move to make space for the error messages.
3. The Порака field should be larger, and there shouldn't be any space between the field and the button.
4. The success Notification is shown and configured correctly. But we have to scroll to see it. It should be fixed. Can you please fix that as well 🙏

edichoska:
@aleksandarmicev 
Screenshot 2024-04-11 at 10 13 08

- Please make sure all input fields are same width.
- I am not sure if the way we are currently showing error messages is good UX.
- Maybe we should have space between the inputs(example 1.5rem, experiment with the value), and in that space we would show the error message.
-In the name field we shouldn't allow any numbers, or special characters except for -. We would need - for any names like Ana-Marija, Mario-Aleksandar 

> The form would be 
> Name
> <name input field>
> <space for showing erors>
> Email
> <email input field>
> <space for showing erors>
> Message
> <textarea>
> <space for showing erors>
> Submit Button

Issue #115, opened 2024-03-23
[FE] Setup React Query


stoimenov95:
**Objective:**

Implement React Query in the application to efficiently manage and cache asynchronous data from server responses. This integration aims to enhance the user experience by reducing loading times and improving data consistency across the app.

**Tasks:**

- Basic Configuration: Set up React Query's QueryClient and provide it at the top level of the application.

- Create a Custom Hook for Fetching Data: Utilize React Query's useQuery hook to fetch data from a test API.

- Error Handling and Loading States: Implement error handling and loading states within the custom hook or component using the data.

- Write a React Query Fetching Example: Create a simple component that uses the custom hook to display data fetched from the API.

**Acceptance Criteria:**

- React Query is correctly installed and configured in the project.

- A custom hook for data fetching is implemented and demonstrated through a sample component.

PR #160, opened 2024-04-12
#115 setup react query

ninman27:
- Set up React Query's QueryClient and provided it at the top level of the app in the Layout
- Created a custom Hook for fetching data
- Implemented error handling and loading states
- Created a simple component Users folder (Users.tsx)

### DIFF
--- a/app/getUsers.tsx
+++ b/app/getUsers.tsx
@@ -1,8 +1,7 @@
 async function getData() {
   try {
     const response = await fetch('https://jsonplaceholder.typicode.com/users');
-    const data = response.json();
-    return data;
+    return response.json();
   } catch (error: any) {
     throw new Error(error);
   }
--- a/app/getUsers.tsx
+++ b/app/getUsers.tsx
@@ -1,0 +1,14 @@
+async function getData() {
+  try {
+    const response = await fetch('https://jsonplaceholder.typicode.com/users');
+    const data = response.json();
+    return data;
+  } catch (error: any) {
+    throw new Error(error);
+  }
+}
+
+export default async function getUsers() {
+  const data = await getData();
+  return data;
+}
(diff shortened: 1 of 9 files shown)


edichoska (dismissed):
@ninman27 can you also include useMutaion?

components/module-components/users/Users.tsx
```diff
@@ -0,0 +1,24 @@
...
+import getUsers from '../../../app/getUsers';
+
+const Users = () => {
+  const { data, isLoading, isError } = useQuery({
+    // eslint-disable-next-line no-return-await
+    queryFn: async () => await getUsers(),
+    queryKey: ['users'],
+  });
+
+  if (isLoading) return <div>Loading</div>;
```
edichoska, line 13: can we display here the global loading component we have?

edichoska:
@ninman27 I keep running into this error:

ninman27:
@edichoska I can't open the image...

edichoska:
Screenshot 2024-04-16 at 00 09 56

app/getUsers.tsx
```diff
@@ -0,0 +1,14 @@
+async function getData() {
+  try {
+    const response = await fetch('https://jsonplaceholder.typicode.com/users');
+    const data = response.json();
```
stoimenov95, line 4: @ninman27 We can return the response.json directly instead of creating the data variable that is only used once.

stoimenov95 (changes requested):

stoimenov95:
> Screenshot 2024-04-16 at 00 09 56

I don't experience this problem...

stoimenov95 (changes requested):
Except for some unnecessary variables, everything else looks awesome @ninman27, great job!

stoimenov95 (approved):
🥳

PR #163, opened 2024-04-14
#76_Contact_Section 

aleksandarmicev:
The form is designed to display an alert if any input is left unfilled. Additionally, there is an alert to indicate successful submission of the form


components/module-components/Contact/circle.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import style from './contact.module.scss';
+
+const Circle: React.FC = () => {
+  return (
```
edichoska, line 5: we need better naming for this component.
maybe SpinningStiker.tsx or SpinningLogoCircle.tsx

edichoska (changes requested):
Screenshot 2024-04-14 at 22 45 58
We should add some styles for when the input is focused, so we dont get the browser default styles.
If there is an error, we can color the border red.
The button should be disabled until all the inputs are properly filed. Maybe the text will be a greyish color.

edichoska:
@aleksandarmicev can we make the inputs on focus to be the same color(the gray one)
also the error messages to only show when we try to submit the form

edichoska (approved):
this looks great 🥇

stoimenov95:
Screenshot 2024-04-16 at 19 42 07
@aleksandarmicev I get this when I try to build the project, can you please take a look?

stoimenov95:
Screenshot 2024-04-16 at 19 43 46
Also, the following errors in the console may not be related to this ticket but I am wondering how are they merged with main @edichoska?

stoimenov95:
> > Screenshot 2024-04-16 at 19 43 46 Also, the following errors in the console may not be related to this ticket but I am wondering how are they merged with main @edichoska?
> 
> @stoimenov95 if you pull main branch do you get this errors? My console is clear. On this branch and on main. I don't get this erorrs.

My bad, main works fine, but this branch is polluted.

stoimenov95:
Screenshot 2024-04-16 at 21 56 04

The text in the input fields should be vertically aligned with the error messages' text. 

It would be nice to increase the send icon a bit to make it more appealing, and I would update the texts to correspond with each other,
e.g., "Внеси го твоето име" - "Задолжително внеси име" or "Внесете го Вашето име" - "Задолжително внесете име", and I would leave out the exclamation points.

stoimenov95 (changes requested):
It would be great if we could separate the styles for the circle in another file - in case we need to reuse them in some other place.
And please take in mind the suggestions I left in the other comment.

You have made great improvements since the initial code, kudos @aleksandarmicev!

edichoska:
Screenshot 2024-04-18 at 10 32 32
@aleksandarmicev Could you please vertically  align the text and the icon.

components/module-components/Contact/ContactForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+    }),
+    onSubmit: async (values, { resetForm }) => {
+      try {
+        const formData: ContactFormData = {
+          first_name: values.username,
+          last_name: '',
+          email: values.email,
+          subject: '',
+          message: values.message,
+          'cf-turnstile-response': '',
```
edichoska, line 44: you will need the changes @Mikivitkovski made, because you need to send here the Cloudflare Turnstile ReCaptcha token. (that PR is still not merged)

components/module-components/Contact/SubmitContactForm.tsx
```diff
@@ -0,0 +1,26 @@
+import axios, { AxiosResponse } from 'axios';
+
+export interface ContactFormData {
+  first_name: string;
+  last_name?: string;
+  email: string;
+  subject?: string;
+  message: string;
+  'cf-turnstile-response'?: string;
+}
```
edichoska, line 10: as you can see in the documentation: https://api.learnhub.mk/docs/#website-POSTcontact
all body parameters are required, so we must provide them.
this will apply only to `cf-turnstile-response` since the backend team will need to remove last_name and subject, since we are not sending that data.

components/module-components/Contact/SubmitContactForm.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type ContactFormResponse = AxiosResponse<any>;
+
+export const submitContactForm = async (
+  formData: ContactFormData
+): Promise<ContactFormResponse> => {
+  try {
+    const response = await axios.post('https://api.learnhub.mk/contact', formData);
```
edichoska, line 18: inlude headers here as well

components/module-components/Contact/SubmitContactForm.tsx
edichoska: sending this req will fail with 422

stoimenov95 (changes requested):
Please resolve comments.

stoimenov95 (approved):
After resolving the comments you can merge this. Awesome work @aleksandarmicev!

Issue #190, opened 2024-04-22
[FE] Create Reusable Blogs Card


edichoska:
- Create reusable Blogs Card that would be used in the Blogs section on the Homepage, and in the Blogs Page.
- Make sure the card/layout is responsive( in the blogs section and blogs page).
- Research how Medium is calculating how much time would it take to read an article and create a function we could use for that in utils.

![Image](https://github.com/learnhubmkd/app/assets/105276163/9bd0d145-d108-4b69-b603-7b6885806923)

PR #201, opened 2024-04-23
#190 - created reusable blogs card

kire21:
- redesigned blog card 

Screenshot from 2024-04-23 19-11-26

Screenshot from 2024-04-23 19-11-11


edichoska (changes requested):
Screenshot 2024-04-23 at 20 49 06
@kire21 

- The title, description and Date | Time spent reading should start at same position on every card.
- The title should be just one line
- Each section should have the same height. And the description and Date | Time spent reading shouldn't move. Example description will have 80px height, it can be 1 line of text or 3 lines of text. That shouldn't affect the sections bellow.

app/page.module.scss
```diff
@@ -1,7 +1,7 @@
 .main {
   display: flex;
   flex-direction: column;
-  justify-content: space-between;
+  // justify-content: space-between;
```
stoimenov95, line 4: Unused comment.

stoimenov95 (changes requested):
I am not sure if I found a key property for each blog card.
Maybe a little zoom should happen for each card on hover, slight box shadow?

Issue #183, opened 2024-04-21
[FE] Create Reusable Input Component


edichoska:
Create reusable input component with appropriate error and success states and styles.

![Image](https://github.com/learnhubmkd/app/assets/105276163/13bbba6f-686e-447b-9b70-e4557a22e2c1)

PR #210, opened 2024-04-26
#183 create reusable input component

Jasminamih:


components/reusable-components/text-input/TextInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const formik = useFormik({
+    initialValues: { [name]: '' },
+    validationSchema: Yup.object({
+      name: Yup.string()
+        .required('Полето е задолжително')
+        .matches(
+          inputRegex,
+          '*Дозволени се следните карактери.... името може да биде на кирилица и латиница'
+        ),
```
edichoska, line 25: This looks good so far.
This input will need to accept:

- validation schema
- placeholder
- error message?
- the input type can be: text, email, number
- name would be also dynamically added
-

edichoska: Also the input should be able to accept props, We might need to use for example: disabled

stoimenov95:
@Jasminamih can we also have an example of the input (implementation for the newsletter maybe) so we can see the design as well?

edichoska (approved):
👏

PR #213, opened 2024-04-28
#185_Created .env and .env.examples files

aleksandarmicev:
Created .env and .env.examples files


.env
```diff
@@ -0,0 +1,2 @@
+NEXT_PUBLIC_API_BASE_URL=https://staging-api.learnhub.mk
+API_KEY=
```
edichoska, line 2: can you double check if we have the .env files in .gitignore
If not please add it.
documentation for reference:
https://git-scm.com/docs/gitignore

aleksandarmicev: I checked .gitignore and the .env file was already added. Is this enough, or should I add another one?

edichoska: So this would ignore files like: .env.production.local
But if we want to have .env ignored we need to add it. 
If it was ignored properly by .gitignore we be able to see it here, and comment on it 🗡️

edichoska (approved):
👏

edichoska (approved):
👏

PR #223, opened 2024-05-06
created perk cards

Jasminamih:


components/module-components/Perk-Cards/PerkCards.tsx
```diff
@@ -0,0 +1,116 @@
...
...9 11 26.5783 10.5786 27.3284 9.82843C28.0786 9.07828 28.5 8.06087 28.5 7C28.5 5.93913 28.0786 4.92172 27.3284 4.17157C26.5783 3.42143 25.5609 3 24.5 3ZM40.506 5C40.112 5 39.7219 5.0776 39.3579 5.22836C38.994 5.37913 38.6633 5.6001 38.3847 5.87868C38.1061 6.15726 37.8851 6.48797 37.7344 6.85195C37.5836 7.21593 37.506 7.60603 37.506 8C37.506 8.39397 37.5836 8.78407 37.7344 9.14805C37.8851 9.51203 38.1061 9.84274 38.3847 10.1213C38.6633 10.3999 38.994 10.6209 39.3579 10.7716C39.7219 10.9224 40.112 11 40.506 11C41.3017 11 42.0647 10.6839 42.6273 10.1213C43.1899 9.55871 43.506 8.79565 43.506 8C43.506 7.20435 43.1899 6.44129 42.6273 5.87868C42.0647 5.31607 41.3017 5 40.506 5ZM8.494 5C8.10004 5 7.70993 5.0776 7.34595 5.22836C6.98197 5.37913 6.65126 5.6001 6.37268 5.87868C6.0941 6.15726 5.87313 6.48797 5.72236 6.85195C5.5716 7.21593 5.494 7.60603 5.494 8C5.494 8.39397 5.5716 8.78407 5.72236 9.14805C5.87313 9.51203 6.0941 9.84274 6.37268 10.1213C6.65126 10.3999 6.98197 10.6209 7.34595 10.7716C7.70993 10.9224 8.10004 11 8.494 11C9.28965 11 10.0527 10.6839 10.6153 10.1213C11.1779 9.55871 11.494 8.79565 11.494 8C11.494 7.20435 11.1779 6.44129 10.6153 5.87868C10.0527 5.31607 9.28965 5 8.494 5Z"
```
edichoska, line 20: Better approach here would be to extract this data into a file called `perkCardsData.ts`
After you have done that, you can rename this component something like PerkCardList, and you would iterate over the data from `perkCardsData.ts` and show PerksCard for each.
This way we will have one place where we have the data, and we can add/remove/edit. And another place where you would have the whole layout with the PerkCards

components/reusable-components/perk-card/PerkCard.module.scss
```diff
@@ -0,0 +1,42 @@
...
+    padding: 16px;
+    border-radius: 8px;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+  }
+}
+.text {
+  font-size: 18px;
+  font-weight: 400;
```
edichoska, line 26: better class naming here:
perkCardWrapper
perkCardTextContent

edichoska (changes requested):

components/module-components/Perk-Cards/PerkCards.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PerkCard from '../../reusable-components/perk-card/PerkCard';
+import style from './PerkCards.module.scss';
+import PerkCardsData from './PerkCardsData'; // Importing the data
```
edichoska, line 4: we don't need the comment here

components/module-components/Perk-Cards/PerkCards.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const PerkCards = () => {
+  console.log(PerkCardsData);
```
edichoska, line 7: remove console.log

components/module-components/Perk-Cards/PerkCards.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <div className={style.perkCardsWrapper}>
+      <div className={style.perkCardsInner}>
+        {PerkCardsData.map((perkCardData) => (
+          <PerkCard
+            key={perkCardData.id}
```
edichoska, line 13: can you create an interface for the PerksCardData and use it here

components/module-components/Perk-Cards/PerkCardsData.ts
```diff
@@ -0,0 +1,33 @@
+const PerkCardsData = [
+  {
+    id: 0,
```
edichoska, line 3: id should not be 0 on all of them

edichoska (changes requested):

components/reusable-components/perk-card/PerkCard.tsx
```diff
@@ -3,7 +3,7 @@ import style from './PerkCard.module.scss';
 
 interface Props {
   perkCardText: string;
-  perkCardIcon: any;
+  perkCardIcon: string;
```
edichoska, line 6: so we should have the same interface here and in components/module-components/Perk-Cards/PerkCards.tsx
you would just export it

RevanchistX: Interfaces should have more descriptive names. Instead use `PerkCardProps`.

Issue #177, opened 2024-04-21
[FE] Mission & Vision


RevanchistX:
![Image](https://github.com/learnhubmkd/app/assets/11035261/a818bf4f-d66b-44df-bb25-b80043c9c5ab)

Note that the red grid pattern should extend 2 rows above the image and 2 rows bellow the image. 
It should be programmatically generated to help with responsiveness.
The example image does not reflect this.

PR #224, opened 2024-05-06
#177 Mission&Vision

aleksandarmicev:


RevanchistX:
@aleksandarmicev 
1. you committed "frames.svg" and "frames.png" but only the png is used. Do we need both images?
2. you added the package "uuid" to dependencies in "package.json", but I don't see it used anywhere. Do we need this?
3. you commented out code in "_colors.scss" - if we don't need this code, delete it.

edichoska:
https://github.com/learnhubmkd/app/assets/105276163/d817488d-bd91-4001-8ffc-a6f390e3855c
Screenshot 2024-05-08 at 21 02 43

The layout looks off on my screen.
The image is distorted. There isn't a margin to separate Mission/Vision from the section above.
The responsiveness on mobile looks okay on iphone 14, but not on the other devices.

@aleksandarmicev

edichoska:
@aleksandarmicev Focus for now on desktop screens to make the image and layout look good and once Dimitar has merged his PR with the fixes for the layout on mobile, you can continue and make improvements.

RevanchistX (changes requested):
Make the change for all images.

edichoska:
Screenshot 2024-05-19 at 01 57 59
Screenshot 2024-05-19 at 01 58 17
Screenshot 2024-05-19 at 01 58 29
Screenshot 2024-05-19 at 01 58 43

On Tablets, the section looks stretched out. You should just add margin block and the text, it shouldn't be 100vh.
- On mobiles we have additional color on top of the section. You may have forgotten to add a background color to the whole section wrapper.
- On smaller desktops it looks like the top and bottom margin are not the same in the Mission/Vision section.
Please take a closer look at the figma file again and compare font sizes, spacings.

edichoska:
Almost done here.
Tablet looks good.
I am seeing some problems on smaller desktops where the section goes outside top and bottom.
On mobile we should have proper padding right.
Screenshot 2024-05-22 at 22 36 45
Screenshot 2024-05-22 at 22 37 00

edichoska:
@aleksandarmicev  Awesome thanks for making the change with the margin.
We need to also remove the padding-top of the Mission Title because we don't need it now, we are controlling the spacing with the margin, and margin + padding results in bigger space on top of the component
Screenshot 2024-05-27 at 14 52 48

edichoska (approved):
🥇 Kuddos Ace, I think we are done here.
You can merge this PR 🥳

edichoska:
@RevanchistX  please take a look, as well 🙏

RevanchistX (approved):
LGTM

PR #227, opened 2024-05-08
206 fe fix layout responsiveness  logo size on mobile

DimitarHristovski:


edichoska:
https://github.com/learnhubmkd/app/assets/105276163/eff11eea-711b-47a0-8725-138837f50363

There is an overflow and it still doesn't look good on mobile.
- You need to add width to the main tag;
- Go through all the sections and check which have a max-width/ min-width added;
- Check the image in the hero section and fix width there

edichoska (approved):
great job @DimitarHristovski  👏

Issue #175, opened 2024-04-21
[FE] Infinite Carousel


RevanchistX:
![Image](https://github.com/learnhubmkd/app/assets/11035261/0ec4b474-9c46-4a99-8195-3d7e8ba15584)

List of elements can be a hardcoded value:
- UI/UX Design
- Frontend Development
- Backend Development
- Quality Assurance
- DevOps
- Marketing
- Product Management

The carousel should:
- be initialized on the left edge, such that no empty background will be visible when the elements start moving.
- infinitely move the elements from the edge of the right screen to the edge of the left screen.
- have no carousel controls.

PR #230, opened 2024-05-09
Infinite carousel #175

kire21:
Created InfiniteCarousel component which is used on the home page.

Screenshot from 2024-05-09 19-54-00

Screenshot from 2024-05-09 19-53-38


components/module-components/infinitie-carousel/InfiniteCarousel.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <p>DevOps</p>
+        </div>
+        <div className={style.slide}>
+          <p>Marketing</p>
+        </div>
+        <div className={style.slide}>
+          <p>Product Management</p>
+        </div>
+        <div className={style.slide}>
+          <p>UI/UX Design</p>
```
edichoska, line 31: Extract this data in another file and map over it

components/module-components/infinitie-carousel/infiniteCarousel.module.scss
```diff
@@ -0,0 +1,68 @@
...
+  background: var(--titles-color);
+  height: 104px;
+  overflow: hidden;
+  position: relative;
+  width: 100vw;
+
+  &::before,
+  &::after {
+    content: '';
+    height: 104px;
```
edichoska, line 30: The height on mobile needs to be 67px

components/module-components/learnhub-animation/LearnhubAnimation.tsx
```diff
@@ -4,8 +4,121 @@ import style from './learnhubAnimation.module.scss';
 
 const LearnhubAnimation = () => {
   return (
-    <div className={style.titleController}>
-      <h1 className={style.titleMessage}>LearnHub.mk</h1>
+    <div className={style.slider}>
```
edichoska, line 7: why do we need this component?

edichoska (changes requested):

components/module-components/infinitie-carousel/InfiniteCarouselData.ts
```diff
@@ -0,0 +1,30 @@
...
+  {
+    text: 'DevOps',
+  },
+  {
+    text: 'Product Management',
+  },
+];
+
+const repeatCount = 4;
+
```
edichoska, line 26: you don't necessarily need this structure of an array of objects
You can create an array of just string values. Then you can then use Array with fill() and flat().
(this is more efficient because we wont be creating new arrays through the spread operator: Array.from({ length: repeatCount }, () => [...InfiniteCarouselData]).flat() ).

edichoska: If you want to keep the structure above, we need to keep in mind how objects are stored.
If we do this multipliedArray[0].text = 'Design'; 
Array.fill()  will make the change in every set because both 'UI/UX' refer to the same object in memory.
Array.from() creates shallow copies, so each repetition in the array will contain a shallow copy of the object, meaning if we do multipliedArray[0].text = 'Design' it will only change the first instance.

edichoska: The data in the array is immutable, meaning we don't want a scenario where the array with the repeated elements is [  {text: 'UI/UX Design',},{text: 'Frontend Development',},{text: 'Backend Development',},{text: 'Quality Assurance',}, {text: 'Marketing',}, {text: 'DevOps',},{text: 'Product Management',},  **{text: 'Design',},**text: 'Frontend Development',},{text: 'Backend Development',},{text: 'Quality Assurance',}, {text: 'Marketing',}, {text: 'DevOps',},{text: 'Product Management',},]

edichoska: you can leave it as it, just wanted to share another perspective ✌️

edichoska (dismissed):
This looks great! Thank you for making the changes 👏

components/module-components/infinitie-carousel/InfiniteCarouselData.ts
kire21: I made the change, you are right. It is the better solution.

edichoska (approved):
⭐

Issue #229, opened 2024-05-09
[FE] Optimize image performance


RevanchistX:
When running `build-dev`, several warnings and suggestions related to images show up.
1. Update the component to resolve this change.
 ![image](https://github.com/learnhubmkd/app/assets/11035261/e3a6fd1f-66c6-42c1-b592-6a4b1eab5ae6)

2. There are some NextJS recommendations with `npm i sharp`, please resolve this as well.
Update: issue appeared due to different yarn versions, sharp is already installed.

RevanchistX:
Hey, @ninman27 

@edichoska mentioned you implemented `sharp` for the project. 
I keep seeing this recommendation: (as mentioned in point 2. of the ticket)
![image](https://github.com/learnhubmkd/app/assets/11035261/55b28ae3-cee1-40d0-bd93-6f742e743b88)

However, after a little bit of investigating, I noticed this only shows up when using `yarn build` and **not** when using `npm run build`.  It seems there's a minor version upgrade `0.33.2 -> 0.33.3` for `sharp`. 

Can you look into this and let me know what you find?
If just the version upgrade solves the problem, that's fine by me.

kire21:
@RevanchistX I tried with yarn build and it is OK. I don't see Next.js's recommendation about the sharp.

![image](https://github.com/learnhubmkd/app/assets/58231146/5da0d219-f126-4e49-b394-33b86a8899d9)

RevanchistX:
Thanks for the feedback @kire21 
It seems we have different yarn versions, but I don't think this is an issue anymore.
I will update the ticket.

PR #232, opened 2024-05-09
#229 optimize image performance

Jasminamih:


components/reusable-components/perk-card/PerkCard.tsx
```diff
@@ -8,10 +9,10 @@ export interface PerkCardsData {
 }
 const PerkCard = ({ perkCardText, perkCardIcon, id }: PerkCardsData) => {
   return (
-    <div className={style.perkCardWrapper}>
-      <div className={style.perkCardInner} key={id}>
+    <div className={style.perkCardWrapper} key={id}>
```
edichoska, line 12: @Jasminamih you need to add key when you map over the perks card data and show the PerksCard
(that would be in PerksData.tsx)

edichoska: remove the key={id} from here

edichoska:
@Jasminamih  Couple of issues with the Perks code

- you should pass the dummy data to the  <PerkCards /> inside page.tsx but into the PerkCards component
- inside the PerkCards component you will create an interface that can be exported (HINT: id will be an optional property) 
- you will get the data here. and iterate over it in the map.
Something like: 
 {perkCardsData.map((perkCardDataItem: PerkCardsDataProps) => (
          <PerkCard
            key={perkCardDataItem.id}
            
 here perkCardsData is the dummy data;  PerkCardsDataProps is the interface
 
 - Then in PerkCard you will just do this const PerkCard = ({ perkCardIcon, perkCardText }: PerkCardsDataProps)

components/module-components/Perk-Cards/PerkCards.tsx
```diff
@@ -1,19 +1,17 @@
...
-interface PerkCardsProps {
-  data: PerkCardsData[];
-}
-const PerkCards = ({ data }: PerkCardsProps) => {
+const PerkCards = () => {
   return (
     <div className={style.perkCardsWrapper}>
       <div className={style.perkCardsInner}>
-        {data.map((perkCardData) => (
+        {PerkCardsData.map((perkCardDataItem) => (
```
edichoska, line 10: you can add the interface here

edichoska (changes requested):
this looks much better, just add the interface here

edichoska:
@Jasminamih look into changing the background color of the section for the dark theme: https://www.figma.com/file/YLlXQlNrfYbBLOyOoIiEp3/learnhub?type=design&node-id=29-450&mode=design

components/reusable-components/perk-card/PerkCard.module.scss
```diff
@@ -11,9 +11,8 @@
   }
 
   .perkCardInner {
-    border: 1px solid #94a3b8;
+    // border: 1px solid #94a3b8;
```
edichoska, line 14: remove comment

edichoska (dismissed):
Thank you for making the changes.
The code looks much cleaner now. 
👏

edichoska (approved):

PR #233, opened 2024-05-09
Created project section

MarijaTrajkoska:


edichoska:
Screenshot 2024-05-14 at 11 39 23
Screenshot 2024-05-14 at 11 39 09
Screenshot 2024-05-14 at 11 30 43

edichoska:
Screenshot 2024-05-14 at 11 46 55
Screenshot 2024-05-14 at 11 47 11
There are some problems on some tablets, and mobile problems with the text

.eslintrc.json
```diff
@@ -23,6 +23,7 @@
     "react/jsx-filename-extension": [1, { "extensions": [".jsx", ".tsx"] }],
     "react/react-in-jsx-scope": "off",
     "react/no-unescaped-entities": 0,
+    "@react/no-array-index-key": "off",
```
RevanchistX, line 26: this rule should not be active, react doesn't work well with index as key.
https://stackoverflow.com/questions/46735483/error-do-not-use-array-index-in-keys/56781201#56781201

instead, try using unique values from the object you're iterating as a key.

edichoska: just revert this. we will use uuid for keys to solve the issue.
We should create separate task for this

edichoska: I think it was okay to use index here, since we wouldn't mutate the array in any way. But safer and better with unique keys.

components/module-components/infinitie-carousel/InfiniteCarousel.tsx
```diff
@@ -7,8 +7,9 @@ const InfiniteCarousel = () => {
   return (
     <div className={style.slider}>
       <div className={style.slideTrack}>
-        {multipliedArray.map((item) => (
-          <div className={style.slide} key={Math.floor(Math.random() * 1000)}>
+        {multipliedArray.map((item, index) => (
+          // eslint-disable-next-line react/no-array-index-key
+          <div className={style.slide} key={index}>
```
RevanchistX, line 12: Since the items are just strings, you can use the item itself as a key.
``` 
{multipliedArray.map((item) => (
          <div className={style.slide} key={item}>
              {item}
          </div>
        ))}
```

components/module-components/ProjectsSection/projectsSection.tsx
```diff
@@ -0,0 +1,64 @@
...
+import useEmblaCarousel from 'embla-carousel-react';
+import style from './projectsSection.module.scss';
+import { useTheme } from '../../../app/context/themeContext';
+import ProjectCard, { ProjectCardProps } from '../../reusable-components/ProjectCard/ProjectCard';
+
+interface Props {
+  cards: ProjectCardProps[];
+}
+
+const ProjectsSection = ({ cards }: Props) => {
```
RevanchistX, line 13: Is this conversion with the interface really necessary?

edichoska:
Screenshot 2024-05-19 at 01 51 56
Screenshot 2024-05-19 at 01 52 36
Here are both cards for comparison.
- We are missing Во Изработка tag both on desktops and mobile 
- Ux/UI Front-end plus the section below with the people should not be displayed on mobile
- These tags have bigger height and width and should have a background color

Please compare your version with the Figma file and make the changes https://www.figma.com/design/YLlXQlNrfYbBLOyOoIiEp3/learnhub?node-id=2723-11614&t=ejvsuI5mAm7URJdl-0
@Jasminamih

edichoska:
Change color of description text(it is a greyish color)
We are missing +3 on the people section (the icons with the people)
Check the Tags spelling and change to: UX/UI
In the Tags section we need to add +3 on all cards. Also +3 should be centrally aligned.
@Jasminamih

edichoska (approved):
Amazing work! Thank you for reworking this until it was perfect.
👏

Issue #179, opened 2024-04-21
[FE] Footer


RevanchistX:
![Image](https://github.com/learnhubmkd/app/assets/11035261/c9ab6f5c-b277-441f-98e9-678f52bbc4d3)

Learnhub Logo  should redirect to home page.
#73 - Enter your Email should be the newsletter component.
#180 - Include the Social Icons component with appropriate icons

PR #236, opened 2024-05-10
#179 footer

Mikivitkovski:


edichoska:
https://github.com/learnhubmkd/app/assets/105276163/9f87be19-2362-4f89-8e15-754f5a2f0ee1

So ideally the inputs wouldn't move when we show the error and success.
It is weird that when we have success the spacing between the inputs increases more.
Also there is some glitch it changes between error and success

edichoska:
General responsive issues on mobile and tablet.
Can we make it in two sections on tablet? If not please just center the content
Screenshot 2024-05-11 at 02 01 34
Screenshot 2024-05-11 at 02 02 02
Screenshot 2024-05-11 at 02 02 17

edichoska:
After successful submission of the newsletter data we should clear the inputs

components/reusable-components/footer/Footer.tsx
```diff
@@ -1,69 +1,173 @@
...
+        body: JSON.stringify(values),
+      });
+      
+      if (response.ok) {
+        setSuccessMessage(true);
+      } else {
+        setErrorMessage(true);
+      }
+      */
+      // Simulating success
```
edichoska, line 60: you can call our API here.
check the submitContactForm for example of how you can do this
the API Documentation:  **https://staging-api.learnhub.mk/docs/**
Since this is still not implemented by the backend team it is good to just extract the function in a separate file and we will test once the backend team is done

components/reusable-components/footer/Footer.tsx
```diff
@@ -1,69 +1,173 @@
...
+      setSuccessMessage(true);
+      toast.success('Успешно испратено!');
+    } catch (error) {
+      setErrorMessage(true);
+      toast.error('Грешка');
```
edichoska, line 65: since we are using toast.error and  toast.success maybe we do not need the Notification component anymore?

components/reusable-components/footer/Footer.tsx
```diff
@@ -1,69 +1,173 @@
...
+                  <Image src={discord} alt="Discord" width={48} height={48} />
+                </Link>
+                <Link
+                  href="https://www.linkedin.com/company/102600044/admin/feed/posts/"
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  <Image src={linkedin} alt="Linkedin" width={48} height={48} />
+                </Link>
+                <Link
```
edichoska, line 136: All this can be a separate component (meaning the icons with the names of the social media)
Maybe we will have a place to reuse it in the future

edichoska (changes requested):
I think we are very close to done on this one.
Overall great job 👏 
Especially great job on changing the instagram icon ⭐

PR #252, opened 2024-05-20
Contact section moves fix #248

### LINKED ISSUE #248: [FE] Contact Section Moves Fix
https://github.com/learnhubmkd/app/assets/105276163/b6ba7949-bd16-4442-a10e-6cf02866f5e1

When the error messages appear the pattern and image moves with it.
Also the inputs inside the form move when an error message appears.

kire21:
Fixed contact section moves


edichoska (approved):
🚀

Issue #246, opened 2024-05-19
[FE] Navigation Style Fix


edichoska:
![Image](https://github.com/learnhubmkd/app/assets/105276163/7b2c24dc-b654-4c28-beaf-0731b3bc8796)

Fix navigation height
Background color of the navigation in the light theme
Test and check the logo(dimensions)

PR #253, opened 2024-05-20
#246 fixed navigation style

ninman27:
Fixed navigation style


edichoska (changes requested):
The navigation still has bigger height than the one in the Figma file

ninman27:
scrshoot1

edichoska (approved):

PR #263, opened 2024-05-20
#254 Simplified regex and form validation

kire21:
Simplified regex and form validation


RevanchistX (approved):
LGTM, blazingly fast regex

PR #265, opened 2024-05-20
182 fe implement storybook

DimitarHristovski:
this is the setup of the storybook for a start... in the future more things will be added.


edichoska:
@DimitarHristovski Why have you uploaded those images?
You shouldn't integrate the storybook as a demo. But with our components.

edichoska:
We are closing this PR, new approach

Issue #266, opened 2024-05-21
[FE] Turnstyle recaptcha library implementation


edichoska:
Research and add implementation of cloudflare turnstile with a library.

Look into:
https://www.npmjs.com/package/@marsidev/react-turnstile (recommended from the official documentation)
https://github.com/marsidev/react-turnstile/tree/main/demos/nextjs
https://www.npmjs.com/package/react-turnstile

PR #271, opened 2024-05-22
#266 turnstile captcha library implementation

Mikivitkovski:


components/reusable-components/turnstile-captcha/CaptchaWidget.tsx
```diff
@@ -1,105 +1,28 @@
...
-  onSuccess?: () => void;
-  onError?: () => void;
-  onExpired?: () => void;
-  siteKey?: string;
-  cData?: string;
-  theme?: 'auto' | 'light' | 'dark';
-  language?: string;
-  tabIndex?: number;
-}
+const TurnstileWidget: React.FC = () => {
```
RevanchistX, line 4: Unnecessary rename, you shouldn't change the function name.

RevanchistX: Additionally: you've modified this entire class but you do not use it anywhere. 
In all other files, you're directly importing `react-turnstile` component, which makes this component obsolete.

components/module-components/contact/ContactForm.tsx
```diff
@@ -34,12 +33,17 @@ const ContactForm = ({ cfTurnstileResponse }: ContactFormProps) => {
         .required('*Пораката е задолжителна'),
     }),
     onSubmit: async (values, { resetForm }) => {
+      if (!turnstileToken) {
+        toast.error('Please complete the captcha');
```
RevanchistX, line 37: Since this is an invisible captcha, it doesn't make sense to ask for the user to complete a captcha which they cannot see.
Change it to something like "Unknown error, please contact administrator" or whatever @edichoska suggests

.env
```diff
@@ -1,2 +1,3 @@
 NEXT_PUBLIC_API_BASE_URL=https://staging-api.learnhub.mk
-API_KEY=
\ No newline at end of file
+API_KEY=
+NEXT_PUBLIC_SITE_KEY=0x4AAAAAAAWq36_j09RgOKQR
```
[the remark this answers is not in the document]
RevanchistX: Additionally: you shouldn't commit `.env` values as they present a vulnerability.

components/reusable-components/footer/Footer.tsx
```diff
@@ -107,6 +125,13 @@ const Footer: React.FC = () => {
           <div className={`title-l ${styles.copyrightSection}`}>
             <p>&copy; 2024 Copyright by LearnHub. All rights reserved.</p>
           </div>
+
+          <Turnstile
+            sitekey={process.env.NEXT_PUBLIC_SITE_KEY || ''}
+            onVerify={(token) => setTurnstileToken(token)}
+            theme="light"
```
RevanchistX, line 132: Even though the captcha is invisible, it should be matching the theme of the website.

components/reusable-components/turnstile-captcha/CaptchaWidget.tsx
RevanchistX: Hint: Delete obsolete code.

RevanchistX (approved):
LGTM

components/reusable-components/footer/Footer.tsx
```diff
@@ -32,13 +39,24 @@ const Footer: React.FC = () => {
...
+  const [turnstileToken, setTurnstileToken] = useState<string | null>(null);
+
   const handleSubmit = async (
     values: { name: string; email: string },
     { resetForm }: FormikHelpers<{ name: string; email: string }>
   ) => {
+    if (!turnstileToken) {
+      toast.error('Please complete the captcha');
+      return;
+    }
```
edichoska, line 51: @Mikivitkovski 
How will the user know what to do in this scenario when the <Turnstile/> is invisible?

cc: @RevanchistX

edichoska: Maybe we can reload the page if there is no token?

RevanchistX: @edichoska will update this in a future scenario, we are going to test the staging environment and the keys for now.

RevanchistX: > Maybe we can reload the page if there is no token?

Reloading the page might lead to infinite reloading if there's something misconfigured. 
The only thing the end user can do is contact the page administrator, which ironically wouldn't work on the contact form.
We can write something like "I guess someone didn't learn how to validate forms hehe" error, if we have to.

edichoska: we should use resetWidget() in this scenario, so the user can try again.
but sure, lets test first.

RevanchistX: @edichoska We tested on Staging and it works as intended.

Issue #249, opened 2024-05-19
[FE] Footer Fixes


edichoska:
https://github.com/learnhubmkd/app/assets/105276163/c74fe26b-88f4-4f5f-ac4e-bf480b02c8c5

The Newsletter section inputs move when error messages appear.
Also on smaller desktops the inline margin should be the same. Now on the right side it looks like there is more space.

PR #274, opened 2024-05-23
#249 footer 2 version for tablet

Mikivitkovski:


components/module-components/contact/ContactForm.tsx
```diff
@@ -78,6 +79,7 @@ const ContactForm = () => {
           type="email"
           field="email"
           formik={formik}
+          inputClass={['']}
```
edichoska, line 82: can't this property be optional? so we don't have to pass empty inputClass

components/reusable-components/button/button.module.scss
```diff
@@ -20,6 +20,10 @@
   }
 }
 
+.smallFooterButton {
```
edichoska, line 23: better class naming here

components/reusable-components/text-input/textInput.module.scss
```diff
@@ -35,6 +36,33 @@
       }
     }
 
+    .newsletterInput {
+      border: 1px solid white;
+      background-color: transparent;
+      color: white !important;
+      height: 3.375rem;
+      -webkit-text-fill-color: #ffff !important;
+      -webkit-text-color: #ffff !important;
```
edichoska, line 45: can we use the white color defined in colors

components/reusable-components/text-input/textInput.module.scss
```diff
@@ -35,6 +36,33 @@
...
+    }
+
+    .lightNewsletterInput {
+      -webkit-box-shadow: inset 0 0 0 30px #121d36 !important;
+      background-clip: content-box !important;
+    }
+
+    .darkNewsletterInput {
+      -webkit-box-shadow: inset 0 0 0 30px #070b15 !important;
+      background-clip: content-box !important;
```
edichoska, line 63: also these colors here, if we don't have them in colors can you please define them there so we can reuse them

edichoska (changes requested):
@Mikivitkovski Miki this looks good. I think this version of the footer is looking much better. We have just some small details to look into it.

components/reusable-components/button/button.module.scss
Mikivitkovski: i didnt find where is used this and i delete it

edichoska (approved):
🚀

Issue #276, opened 2024-05-25
[FE] Implement subscribe to a newsletter


edichoska:
Look into the API Documentation and implement the subscribe functionality.

PR #277, opened 2024-05-26
#276 implement subscribe to a newsletter

Mikivitkovski:


package.json
```diff
@@ -14,6 +14,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "@marsidev/react-turnstile": "^0.6.1",
     "@tanstack/react-query": "^5.29.2",
```
edichoska, line 18: @Mikivitkovski  are we using both libraries? If yes why?

components/reusable-components/footer/SubmitNewsletterForm.tsx
```diff
@@ -1,28 +1,29 @@
...
-        });
-      
-        if (!response.ok) {
-          throw new Error(`HTTP error! status: ${response.status}`);
-        }
-        return response.json();
-        */
+
+export const submitNewsletterForm = async (formData: NewsletterFormData): Promise<string> => {
+  const url = new URL('https://staging-api.learnhub.mk/newsletter-subscribers');
```
edichoska, line 8: this should go into the .env file

components/reusable-components/footer/Footer.tsx
```diff
@@ -41,28 +41,29 @@ const Footer: React.FC = () => {
...
+      first_name: values.name,
+      email: values.email,
+      'cf-turnstile-response': turnstileToken,
     };
 
     try {
-      await submitNewsletterForm(formValues);
+      const responseMessage = await submitNewsletterForm(formValues);
+      if (responseMessage === '') {
+        toast.info(responseMessage);
```
edichoska, line 58: why would be display toast info?

edichoska:
Screenshot 2024-05-26 at 20 27 22
@Mikivitkovski when we type the text is white, when i paste a text it is the correct color.

edichoska (commented):
You should check if the form is successfully submited. 
if yes we display the toast.success with the message, if not we display the toast error
Also disable the button till the user has entered the two fields, since both are required.

edichoska (approved):
⭐

Issue #291, opened 2024-05-28
[FE] Change in margin and padding in the Mission/Vision Card


edichoska:
![Image](https://github.com/learnhubmk/app/assets/105276163/c5879a1e-d7da-4325-a046-facf1ca64c69)

Check the Figma File for reference: https://www.figma.com/design/YLlXQlNrfYbBLOyOoIiEp3/learnhub?node-id=3213-9905&t=MPhJgSCGD2ZPQBl3-0
There is a change in the margin and padding of the card, please compare with the card we have implemented and replicate the Figma design.

PR #294, opened 2024-05-28
#291 change in margin and padding in the mission/vision card

Mikivitkovski:


edichoska (approved):
Lets merge this @Mikivitkovski and we will check again if there are any changes in the design.

Issue #287, opened 2024-05-28
[FE] Update color and gradient in Mission/VIsion


edichoska:
The color of Mission/Vission card in dark theme is changed.
There is also a change in the gradient in dark theme of the section.
Reference: https://www.figma.com/design/YLlXQlNrfYbBLOyOoIiEp3/learnhub?node-id=29-450&t=MPhJgSCGD2ZPQBl3-0

PR #295, opened 2024-05-28
#287 Update color and gradient in Mission/Vision

aleksandarmicev:
Changed color for: 1.Gradient, 2. Color of Mission/Vision text card


components/module-components/Mission&Vision/missionVision.module.scss
```diff
@@ -139,7 +135,7 @@
 }
 
 .missionTitleDark {
-  background: rgba(54, 67, 98, 1);
+  background: var(--Primary, hsla(222, 50%, 14%, 1));
```
edichoska, line 138: could you update the colors file as well? (colors.scss) and add this values as well so we can reuse them through out the project

app/styles/globals/_colors.scss
```diff
@@ -19,61 +19,4 @@
...
-  $clr-secondary-light-mode: hsl(31, 94%, 86%);
-  $clr-tertiary-light-mode: hsl(222, 50%, 14%);
-  $clr-text-light-mode: hsl(215, 19%, 35%);
-  $clr-elevated-elements-light-mode: hsl(210, 40%, 98%);
-  $clr-titles-light-mode: hsl(222, 50%, 14%);
-  $clr-tertiary-button-light-mode: hsl(0, 0%, 0%);
-
-  $clr-primary-dark-mode: #121d36;
-  $clr-secondary-dark-mode: hsl(222, 19%, 44%);
-  $clr-tertiary-dark-mode: hsl(222, 52%, 5%);
```
edichoska, line 33: @aleksandarmicev  Ace is this a mistake or on purpose? I see you have deleted all the colors from colors.scss

aleksandarmicev: It's not a mistake, I purpose in the code to exist separate file for color variables, and in color to be only primary colors

edichoska: why would we do that?

edichoska: Please undo that.
This ticket is only for updating the colors used in the Mission/Vision section.
We can discuss updating the Sass file structure. Maybe we can use the 7:1 method. But we should create a separate issue for that. 🙏 
Try to keep the PRs only about what was mentioned in the ticket.

aleksandarmicev: Ok, I'll revert these changes.

edichoska: Thank you

edichoska (approved):
🚀

Issue #290, opened 2024-05-28
[FE] Center text in FAQ Section


edichoska:
![Image](https://github.com/learnhubmk/app/assets/105276163/5361dcf7-537f-4345-832b-bb84138d0a94)

The title and subtitle text in the FAQ section should be centered. 
Figma file for reference: https://www.figma.com/design/YLlXQlNrfYbBLOyOoIiEp3/learnhub?node-id=3213-9905&t=MPhJgSCGD2ZPQBl3-0

PR #297, opened 2024-05-29
#290 center text faq section

MarijaTrajkoska:


components/module-components/faq/FAQ.module.scss
```diff
@@ -54,21 +54,35 @@
...
+  text-align: center;
 
   @include u.breakpoint(medium) {
     margin-bottom: 1.5rem;
   }
+
+  @include u.breakpoint(xsmall) {
+    font-size: 2.5rem;
+    text-align: start;
+  }
```
edichoska, line 67: can you please double check this values for the font size?
I can see that the font-size on mobile is 28(1.75rem) and on desktop 40px(2.5rem)
look into the typography file and check is you could reuse something from there?
See how font size is defined in other sections in order to make it responsive
@MarijaTrajkoska

edichoska (changes requested):

edichoska (approved):
Looks good. Thank you for making the changes @MarijaTrajkoska

PR #312, opened 2024-06-09
Main update

edichoska:


components/module-components/projectsSection/projectData.ts
RevanchistX, line 1: Replace `lorem ipsum` for something more substantial.

RevanchistX (changes requested):
LGTM, just change the dummy data

components/module-components/projectsSection/projectData.ts
edichoska: @RevanchistX someone else from the team will update the remaining lorem ipsum text once we come up with descriptions. This is so we update main

davorminchorov: I would suggest that we update the text first before releasing the new design.

RevanchistX (approved):

Issue #314, opened 2024-06-10
[FE] Contact overlap bug on tablets


edichoska:
![Image](https://github.com/learnhubmk/app/assets/105276163/30f6f3ea-8667-4811-a456-305fa3a6606c)

PR #317, opened 2024-06-10
#314 contact overlap bug on tablets

Mikivitkovski:


edichoska (approved):
Thanks for the quick fix @Mikivitkovski 👏

Issue #325, opened 2024-06-11
[FE] Boarder on the Mission and Vision card - Dark mode web


sanjadivanisova:
Please remove the border from the Mission and Vision card in dark mode for web. There is currently a white border on the card. 

![Image](https://github.com/learnhubmk/app/assets/161955046/0553c26b-18a6-420b-82b7-96ffc565734e)

For reference, please look at Figma: 

https://www.figma.com/design/YLlXQlNrfYbBLOyOoIiEp3/learnhub?node-id=2496-3225&t=1cxD4Q8PJUVdjhej-0

edichoska:
#326 @aleksandarmicev  you can do both of these on one PR

PR #330, opened 2024-06-11
#325-Remove border on M&V card

aleksandarmicev:
#325-Remove border on M&V card


edichoska (changes requested):
why are MissionVission.tsx and missionVision.module.scss added here again?
shouldn't this PR be only about removing the border

aleksandarmicev:
> why are MissionVission.tsx and missionVision.module.scss added here again? shouldn't this PR be only about removing the border

Some problem with VS CODE. Also Vercel — Deployment has failed, I don't know why, will check now

Issue #326, opened 2024-06-11
[FE] Boarder on the Mission and Vision card - Light mode web


sanjadivanisova:
Please remove the border on the Mission and Vision card in light mode for web.  Additionally, please change the colour of the card itself to: 

#FAFAFC

This is the current deployed version: 

![Image](https://github.com/learnhubmk/app/assets/161955046/979d4ed2-8b52-487d-b059-f8adcb10ff94)

This is the design in Figma:

![Image](https://github.com/learnhubmk/app/assets/161955046/f96c3752-2f53-4d46-be29-08e80cde03ad)

Figma file: 

https://www.figma.com/design/YLlXQlNrfYbBLOyOoIiEp3/learnhub?node-id=29-450&t=1cxD4Q8PJUVdjhej-0

PR #333, opened 2024-06-11
#325 and #326 Remove border + changed color in light mode 

aleksandarmicev:
Remove border on the M&V card, and also changed color in light mode according to figma


edichoska (approved):
👏

Issue #328, opened 2024-06-11
[FE] Discrepancy between the two cards under the Projects section - Light and Dark mode web 


sanjadivanisova:
There currently is a discrepancy between the two cards under the Projects section for both Light and Dark mode for web in terms of padding (between the image and the card title). 

Current deployed design: 

![Image](https://github.com/learnhubmk/app/assets/161955046/25468e22-6349-4a14-baa9-35bcd2d87869)

Please change the padding of the LearnHub card to correspond with the padding of the LinkMe card. 

Design reference in Figma: 

https://www.figma.com/design/YLlXQlNrfYbBLOyOoIiEp3/learnhub?node-id=2496-3225&t=1cxD4Q8PJUVdjhej-0

Furthermore, we need to change the data in terms of categories and number of users in the cards so that it is different in both cards. Please speak with Emilija for further reference. 

**Summary from Emilija:** 

On bigger desktops 1920px and above, the cards look different. The first one is bigger, also the title and description move.
Additionally lets change the dummy data:
In LH Project card add +22 in the people section
In LinkMe Project card + 12 in the people section

In LH Project Card add another tag Backend

PR #339, opened 2024-06-12
#328  Update Gap, margin Bottom, card to be identical size, change an…

Mikivitkovski:
…d add to dummy data tagNumber and participants, and update tagNumber and participants from figma, and change place for backend tag in LH project


edichoska (approved):
🚀

Issue #315, opened 2024-06-10
[FE] Create Reusable Modal Component


edichoska:
- [ ] Create reusable Modal Component that should contain:
- X button 
- Title
- Description(this would be optional)
- Two Buttons( Cancel and Action Button(we should be able to pass the text as a prop))
- (the action button would be in the primary color, cancel in secondary)

- [ ] The component should open on click of a button (any button in the app)
- [ ] On click outside of the modal we should be able to close it. 
- [ ] Also we should be able to close the modal on click of X button
- [ ] The page should have an overlay, on top of which we would display the modal
- [ ] We shouldn't be able to scroll the page when the modal is open

PR #347, opened 2024-06-13
Reusable modal

Mario-k4:
#315

### DIFF
--- a/app/admin-dashboard/page.tsx
+++ b/app/admin-dashboard/page.tsx
@@ -1,18 +1,14 @@
 'use client';
 
-import React, { useState } from 'react';
+import React from 'react';
 import ReusableModal from '../../components/reusable-components/reusable-modal/ReusableModal';
+import {
+  ModalProvider,
+  useModal,
+} from '../../components/reusable-components/reusable-modal/ModalContext';
 
 const AdminDashboard = () => {
-  const [isModalOpen, setIsModalOpen] = useState(false);
-
-  const showModal = () => {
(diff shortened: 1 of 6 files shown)


app/admin-dashboard/page.tsx
```diff
@@ -1,5 +1,32 @@
...
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const showModal = () => {
+    setIsModalOpen(true);
+  };
+
+  const closeModal = () => {
+    setIsModalOpen(false);
+  };
+
```
edichoska, line 16: since we will be using the modal in multiple components, we would need to write this logic in all of them.
Lets try to make this more reusable.
(hint: use context)

app/admin-dashboard/page.tsx
```diff
@@ -1,5 +1,32 @@
...
+  return (
+    <div>
+      <button type="button" onClick={showModal}>
+        Open Modal
+      </button>
+      <ReusableModal
+        title="Modal Title"
```
edichoska, line 23: currently on the modal i cannot see a title, maybe you made the text white?

edichoska (changes requested):
When we open the modal and show the overlay, the scrollbar is hidden, when we click on the overlay to close the modal, the whole layout shifts because the scrollbar appears.
you need to find a way to acomodate for the scrollbar width

edichoska (approved):
Great job @Mario-k4  👏

Issue #369, opened 2024-06-16
[FE] Update the images in Project Section


edichoska:
Update the current desig with the images from Figma.
Change made because we have duplicate images.

![Image](https://github.com/learnhubmk/app/assets/105276163/c3180546-0438-4df7-a8ef-bec566c92da3)

PR #370, opened 2024-06-17
replaced the images in the project section

Mario-k4:
#369


edichoska:
@Mario-k4 can you remove the image that was previously used?

edichoska (approved):

PR #379, opened 2024-06-19
Test meta logo how it looks published

Mikivitkovski:


edichoska:
@Mikivitkovski make branch from main, and the PR base should be main. We shouldn't add all this components to main for now

Mikivitkovski:
> @Mikivitkovski make branch from main, and the PR base should be main. We shouldn't add all this components to main for now

Its on main i change that

Issue #301, opened 2024-05-30
[FE] Create Sign Up Page Design


edichoska:
- [ ]  Create SignUp route
- [ ]  Display SignUp page design
- [ ] Create reusable form design for the SignUp form, and create the layout as shown in the Figma File: (option 1)

 https://www.figma.com/design/YLlXQlNrfYbBLOyOoIiEp3/learnhub?node-id=29-450

Note: This ticket is only for routing and for implementing the design

PR #401, opened 2024-06-21
sign up page design with mobile responsiveness

Mario-k4:
#301 Not final,still mising design for tablet


edichoska (commented):
Since both the login and the sign up page have similar layout, please find a reusable solution that could work for both of them.
@Jasminamih collaborate with Mario on that one.

components/reusable-components/reusable-form/ReusableForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+        />
+        <TextInput
+          placeholder="Внесете ја вашата лозинка"
+          type="password"
+          label="Лозинка *"
+          name="password"
+          field="password"
+          formik={formik}
+          inputClass={['signUpInput']}
+          onChange={handlePasswordChange}
```
edichoska, line 116: First issue here is how we are handling onChange in the custom TextInput component, right now it handles only     `formik.handleChange(e);` but it should also handle the scenario where we are passing a onChange function like here

components/reusable-components/reusable-form/ReusableForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  const formik: FormikProps<{
+    firstName: string;
+    lastName: string;
+    email: string;
+    password: string;
+  }> = useFormik({
+    initialValues,
+    validationSchema,
+    onSubmit: () => {},
```
edichoska, line 55: Here on Submit we will update that state by testing each condition . Something like:
  ```
  onSubmit: (values) => {
      setPasswordValidation({
        uppercase: /[A-Z]/.test(values.password),
        specialChar: /[!@#$%^&*(),.?":{}|<>]/.test(values.password),
        minLength: values.password.length >= 8,
      });
    },
```

File: components/reusable-components/reusable-form/ReusableForm.tsx [hunk 3]
```diff
@@ -0,0 +1,192 @@
...
+      return /[A-Z]/.test(password);
+    }
+    if (rule === 'specialChar') {
+      return /[!@#$%^&*(),.?":{}|<>]/.test(password);
+    }
+    if (rule === 'minLength') {
+      return password.length >= 8;
+    }
+    return false;
+  };
```
edichoska, line 75: we won't need this.

components/reusable-components/reusable-form/ReusableForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+    initialValues,
+    validationSchema,
+    onSubmit: () => {},
+  });
+
+  const handlePasswordChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = e.target;
+    setPassword(value);
+    formik.handleChange(e);
+  };
```
edichoska, line 62: Here we will update the new state as well.
Here you will test the conditions and merge the new state with the previous state.
When testing the contitions you will check if the condition is met it will be true, if not it would be null. We need the null value here since the icon color should stay grey if the condition is not met. We are changing it to red on Submit.


About changing the icon color.
Since we have an svg icon, we can create a component that would have a color prop and will return the svg icon.
You will make the following change in the svg `stroke={color}`
and also the prop would have a default value(the gray color).

When we use that component here, we would have something like: 
`  <CheckPasswordValidityIcon color={getColor(valid)} />`

edichoska: You need a function that  will return a hex value, it will have an argument that can be boolean or null
if null return the grey color if it is true return the green if not the red

components/reusable-components/reusable-form/ReusableForm.tsx
```diff
@@ -0,0 +1,192 @@
...
+              src={check}
+              alt="check"
+              className={
+                isPasswordValid('specialChar', formik.values.password) ? style.valid : style.invalid
+              }
+            />
+            <p>Еден специјален знак или симбол</p>
+          </div>
+          <div className={style.requirementsItem}>
+            <Image
```
edichoska, line 140: here we can map over the entries of the new state

edichoska (changes requested):
@Mario-k4 try to complete this with the directions i gave you in the comments. If you have any problems let me know and we can finish it together.

components/reusable-components/reusable-form/ReusableForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+const ReusableForm = () => {
+  const { theme } = useTheme();
+  const lightTheme = theme === 'light';
+
+  const [password, setPassword] = useState<string>('');
+  const [passwordValidation, setPasswordValidation] = useState<{
+    uppercase: boolean | null;
+    specialChar: boolean | null;
+    minLength: boolean | null;
+  }>({
```
edichoska, line 30: create interface for these values

components/reusable-components/reusable-form/ReusableForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+          {Object.entries(passwordValidation).map(([rule, valid]) => (
+            <div className={style.requirementsItem} key={rule}>
+              <CheckPasswordValidityIcon color={getColor(valid)} />
+              <p>
+                {rule === 'uppercase'
+                  ? 'Една голема буква'
+                  : rule === 'specialChar'
+                    ? 'Еден специјален знак или симбол'
+                    : 'Минимум 8 карактери'}
+              </p>
```
edichoska, line 147: Here you can define an object ruleMessages where each key of the object corresponds to a rule, and the value is the associated message.
You would use it here like this:
<p>{ruleMessages[rule] || ruleMessages.default}</p>

components/reusable-components/reusable-form/ReusableForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+            </div>
+          ))}
+        </div>
+
+        <label htmlFor="checkbox">
+          <input type="checkbox" id="checkbox" />
+          Се согласувам со сите Terms & Conditions
+        </label>
```
edichoska, line 155: This needs to be required. The user shouldn't be able to Sign Up without checking this.

app/admin-dashboard/[slug]/page.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { useSearchParams } from 'next/navigation';
+
+const UsersPage = () => {
+  const searchParams = useSearchParams();
+
+  const search = searchParams.get('id');
+  return <div>Users: {search}</div>;
+};
+
+export default UsersPage;
```
Comment (edichoska) [inline app/admin-dashboard/[slug]/page.tsx:12]: Not sure why these changes are on this PR. Please revert the change as it is not applicable here.

edichoska (approved):
💅

Issue #409, opened 2024-06-29
[FE] Bug: Theme switch


edichoska:
The theme switch is not working as intended on the login route(possibly other routes as well).
When i change the theme and refresh the page it doesn't remember the choice and reverts to the default state.

Expected behaviour: The user changes the theme, reloads page, we remember the choice and the theme persists

PR #410, opened 2024-07-01
#409-Theme-Switch

aleksandarmicev:
Made changes in ThemeContext.tsx and implement Cookies for solving problem with the theme switch.


edichoska:
@aleksandarmicev we should test this properly on staging first. Please resolve conflicts and lets merge this and test. Then we can release this fix to main.

aleksandarmicev:
> @aleksandarmicev we should test this properly on staging first. Please resolve conflicts and lets merge this and test. Then we can release this fix to main.

I checked this PR and I think that because there is a cookie, running CMD 'npm i' result for eliminating errors. I think the other parts are ok.

Please, let me know if there is a necessary for further adjustments!

davorminchorov:
If you want to test the code on staging, it may be better if you create a `-staging` branch and merge that branch to staging so that this branch remains clean and gets merged into main directly without the need to do any cherry-picking.

aleksandarmicev:
> > If you want to test the code on staging, it may be better if you create a `-staging` branch and merge that branch to staging so that this branch remains clean and gets merged into main directly without the need to do any cherry-picking.
> 
> we don't need this on main. the toogle is working fine on main. there is an issue on staging with the new pages. So the PR should be made towards staging, and we will see if it solves the problem.



Conflict resolved, this PR will go to the staging, according this discussion

edichoska:
@aleksandarmicev Just tested this, works! 🥳  Thanks for the quick fix!

edichoska (approved):

PR #417, opened 2024-07-06
Signup and login reusable layout

Jasminamih:


components/reusable-components/signup-and-login-layout/SignupAndLoginLayout.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import React from 'react';
+import Image from 'next/image';
+import styles from './SignupAndLoginLayout.module.scss';
+import { useTheme } from '../../../app/context/themeContext';
+
+interface SignupAndLoginProps {
+  welcomeTitle: string;
+  welcomeSubtitle: string;
+  // form?: JSX.Element;
```
edichoska, line 11: We will accept children here. It can be optional.

components/reusable-components/login-form/LoginForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+          type="email"
+          field="email"
+          formik={formik}
+          isRequired
+        />
+        <TextInput
+          placeholder="Внесете го вашиот пасворд"
+          label="Пасворд"
+          name="password"
+          type="password"
```
edichoska, line 85: Why is the Login form on this PR?
This PR should only be about the shared layout the login and sign up share. Nothing else.
Additionally, I believe we already have the patterns. Aren't these the same patterns we already use in Mission and Vision?

edichoska:
@Jasminamih There is a mismatch between the package.json and package.lock.json
jquery@3.7.1 is missing from the lock file

Issue #386, opened 2024-06-19
[FE] Add tags component


edichoska:
On Add we should open a search component similar to this example on Linkedin. We could add tags only from existing ones, or we can add tags that we dont have already. If the tag doesnt exist a button Create Tag would appear and we would create and add the tag.
![Image](https://github.com/learnhubmk/app/assets/105276163/bc8c8095-1285-4cf7-a198-62ef3fa4db8a)
![Image](https://github.com/learnhubmk/app/assets/105276163/d5eec8e7-9705-4cf5-83c4-9d6e04748a74)

PR #426, opened 2024-07-17
#386 add tags component

Jasminamih:


components/reusable-components/add-skill/AddSkills.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <TagsModal
+          existingTags={existingTags}
+          selectedTags={selectedTags}
+          setSelectedTags={setSelectedTags}
+          onClose={closeModal}
+        />
+      )}
+      <div className={styles.selectedTags}>
+        {/* eslint-disable jsx-a11y/no-static-element-interactions */}
+        {/* eslint-disable jsx-a11y/click-events-have-key-events */}
```
edichoska, line 40: This eslint errors are easy to fix lets do that instead of disabling the rules:
1. Convert the span element to button
2. Add role to the button
3. Add aria-label = 'Remove' or aria-label='Delete'
more information on why we need to add aria-labels to help with assistive technology: https://stackoverflow.com/questions/22039910/what-is-aria-label-and-how-should-i-use-it

The problem with using span here is that it is non-native interactive element. We need to make sure we are using semantic html.

edichoska (changes requested):
The functionality is working as intended. I also like that we are able to remove the element from the modal and it removes it from the screen as well.
Just the small eslint fixes and we are good.

components/reusable-components/tags-modal/TagsModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    if (query) {
+      const filtered = existingTags.filter((tag) =>
+        tag.toLowerCase().includes(query.toLowerCase())
+      );
+      setFilteredTags(filtered);
+    } else {
+      setFilteredTags([]);
+    }
+  };
```
edichoska, line 32: We should use debounce on the search input. So we can improve performance and avoid sending requests on every keystroke.

edichoska: The more I think about it, this should be implemented as following.
We will fetch the tags list, and memoize it. Search through that list, if we don't find that item we can send a request to add the tag.

components/reusable-components/tags-modal/TagsModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (!selectedTags.includes(tag)) {
+      setSelectedTags([...selectedTags, tag]);
+    }
+    setSearchTag('');
+    setFilteredTags([]);
+  };
+
+  const removeTag = (tag: string) => {
+    setSelectedTags(selectedTags.filter((item) => item !== tag));
+  };
```
edichoska, line 44: To avoid issues with stale state, we should use  the snapshot of the most recent state. something like:   setSelectedTags((prevTags) ....

components/reusable-components/tags-modal/TagsModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  return (
+    <div className={styles.modalWrapper}>
```
edichoska, line 47: Use the reusable modal here

components/reusable-components/tags-modal/TagsModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <div className={styles.modalContent}>
+        <button type="button" className={styles.closeButton} onClick={onClose}>
+          <i aria-label="Close button" className="bi bi-x" />
+        </button>
+        {/* eslint-disable jsx-a11y/label-has-associated-control  */}
```
edichoska, line 52: lets try using this in the eslint configuration:
{
  "rules": {
    "jsx-a11y/label-has-associated-control": [
      "error",
      {
        "assert": "htmlFor",
        "depth": 3
      }
    ]
  }
}

edichoska (changes requested):

components/reusable-components/tags-modal/TagsModal.tsx
edichoska: Its trying to enfore us to use it like:
<label><input/> </label>
using htmlFor with same value as the id should fix that, but since it is still shouting at us, lets update it as I mentioned above.

edichoska (changes requested):
Can you attach a screenshot of how the UI looks? And also implement it so we can test it if it works as intended.

edichoska (approved):

Issue #382, opened 2024-06-19
[FE] Implement Authentication 


edichoska:
- [x] Add Sign In and Sign Up buttons in Navigation
- [x] Create authentication context and add login, logout, fetch user methods
- [x] Save token and create session

davorminchorov:
We will be using Laravel Sanctum for authentication, specifically the SPA mode.

stefanatanasovmk:
Some questions about the ticket:

1. There is 2 login routes in the API, one for `content` and the other for `admin`. Shall we do the login only for `content` users?
2. Shall we fetch the `XSRF-TOKEN` route on initial page load, as I presume the token needs to be included when making the login request?
3. Is this expected workflow: user enter password and email -> client make an request to the server `/login` route -> the server returns token and the user info -> client encrypt the token and user info, stringify it and save it into a `session` cookie? Or does the server sets a cookie when login is successful?
4. What the login route returns when user is authenticated, it's not specified in the docs.
5. Shall me make the login for the third-party auth as well, as I saw there is route for github, google, linkedin... login? 
6. Can we get already registered user, or where to signup? As I didn't see the signup route in the docs. and I would like to see how it works when log in.

davorminchorov:
1. There are 3 endpoints for logging in currently: admin, content and platform. This ticket is for logging in to the content endpoint I assume, although all 3 endpoints will have a similar implementation.

As for the process on how to authenticate, you can [check the Laravel Docs](https://laravel.com/docs/11.x/sanctum#spa-authenticating) to understand the process.

The 3rd party logging in (and sign up) is for the platform only. There are no sign up routes for content and admin, they are supposed to be invite only.

I can provide you content manager credentials for signing in.

edichoska:
@davorminchorov Could the BE team improve the docs about 4.?
As Davor explained:
We don't have Sign Up for the  content panel. The user will be invited by email. On click on that link the user will be redirected to a page, where they would just add a password & confirm that password.( I assume this is still not implemented? @davorminchorov)
1. Yes. Focus only on content panel.
2 & 3. On the right track, read documentation.
4. Will see with BE team.
5. We don't have SSO for the content panel.

davorminchorov:
The current implementation does not support the SPA mode yet, but the docs will be updated once that's done.

The invitation of users will be done via the admin panel so that's irrelevant for this ticket specifically.

edichoska:
Of course it is relevant for Stefan to know how the whole flow is going to be.

[...]

PR #428, opened 2024-07-22
#382 implement authentication

stefanatanasovmk:


edichoska (changes requested):
Is this tested and working? @stefanatanasovmk @MarijaTrajkoska

edichoska:
Please resolve conflicts before requesting a review 🙏 
Also it is a good practice to run npm run build-dev to catch the errors that can occur during vercel deployment
This is the Error:
Screenshot 2024-07-23 at 15 44 12

utils/actions/session.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function setSession(session: Session): Promise<void> {
+  const cookie = JSON.stringify(session);
+  cookies().set('session', cookie, {
+    expires: Date.now() + 1000 * 60 * 60 * 24 * 7, // 7 days
+    httpOnly: false,
+  });
+}
+
+export async function getSession(): Promise<Session | null> {
+  const session = cookies().get('session');
```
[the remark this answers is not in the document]
stefanatanasovmk: We don't need both, removed auth.ts

components/login.tsx
```diff
@@ -0,0 +1,55 @@
...
+  useEffect(() => {
+    const getCsrf = () =>
+      fetch('https://staging-api.learnhub.mk/sanctum/csrf-cookie', {
+        headers: {
+          Accept: 'application/json',
+          Referer: 'learnhub.mk',
+        },
+        credentials: 'include',
+      });
+    (async () => {
```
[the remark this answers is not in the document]
stefanatanasovmk: No, removed.

app/context/authContext.tsx
```diff
@@ -0,0 +1,175 @@
+'use client';
+
+import React, { createContext, useContext, ReactNode, useMemo } from 'react';
+import { MutationStatus, QueryStatus, useMutation, useQuery } from '@tanstack/react-query';
+import { clearSession, getSession, setSession } from '../../utils/actions/session';
+
+const baseUrl = process.env.NEXT_PUBLIC_API_BASE_URL || 'http://localhost:8000';
```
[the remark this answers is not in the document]
stefanatanasovmk: Done !

stefanatanasovmk:
The checks pass now. Some unused components which were committed couple of days ago, I forgot to remove with the latest final changes yesterday.

app/context/authContext.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+const baseUrl = process.env.NEXT_PUBLIC_API_BASE_URL || '';
+
+export type Role = 'admin' | 'content_manager' | 'member';
+
+type UserType = {
+  id: number | string;
+  is_verified: boolean;
+  email: string;
+  role: Array<Role>;
```
davorminchorov, line 15: This was an array initially but we changed it to be a string for now. Only one role per user will be used for the time being. It may be changed in the near future if requirements change.

edichoska: @stefanatanasovmk

edichoska: we should add role automatically, depending from where the user logs in

app/context/authContext.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const {
+    status: logoutStatus,
+    mutate: logout,
+    error: logoutError,
+  } = useMutation({
+    mutationKey: ['logout'],
+    mutationFn: async () => {
+      const session = await getSession();
+      if (!session?.token || !session.role) throw new Error('No session found');
+      await fetch(`${getUrl(session.role)}/logout`, {
```
edichoska, line 123: we can create a middleware that would check the session before  each request

edichoska:
we are also missing the refresh token mechanism implemented

app/context/authContext.tsx
stefanatanasovmk: @davorminchorov This is a `UserType`, I'll change the role to be of type `Role` and not an array.
@edichoska The user role is fetched from the backend. When login in, the `login` function expect an argument `role` because the backend have different routes for login depending of the user role. But, I assume that the user role will be assigned on registration on the backend, as there is different urls for login, I guess there will be different urls for registration depending of the user's role.

app/context/authContext.tsx
stefanatanasovmk: @edichoska Shall we do something like here: https://github.com/laravel/breeze-next/blob/master/src/hooks/auth.js

In this example in the `useAuth` hook we pass `middleware` prop that decide if the user needs to be authenticated, and if so we check if the `getUserStatus` from the `AuthContext` is `success` and if so we don't do anything, otherwise we redirect the user to the login route. 
This can create protected routes, everywhere we need the user to be logged in, we can call `useAuth` with passed `middleware: auth` as prop.

Example:

```
export const useAuth = ({
  middleware,
  redirectIfAuthenticatedTo,
}: {
  middleware: 'auth' | 'guest';
  redirectIfAuthenticatedTo: string;
}) => {
  const context = useContext(AuthContext);
  const router = useRouter();
  if (context === undefined) {
    throw new Error('useAuth must be used within an AuthProvider');
  }
  const { getUserStatus } = context;

  useEffect(() => {
    if (middleware !== 'auth') return;
    if (getUserStatus === 'pending') return;
    if (getUserStatus === 'error') {
      router.push('/login');
    }
    if (getUserStatus === 'success' && redirectIfAuthenticatedTo)
      router.push(redirectIfAuthenticatedTo);
  }, [getUserStatus, middleware, redirectIfAuthenticatedTo, router]);
  return { ...context };
};

```

This is all in the client-side and don't have anything to do with the nextjs middleware. But in the nextjs middleware, we can check to see the expiration date of the session, and refresh it if it is expiring in a certain time?

stoimenov95: @stefanatanasovmk, from my side, we can go with the proposed solution, and integrating server-side checks with Next.js middleware for session expiration and refresh could provide a more robust solution.

app/context/authContext.tsx
```diff
@@ -0,0 +1,204 @@
...
+import { useRouter } from 'next/navigation';
+import { MutationStatus, QueryStatus, useMutation, useQuery } from '@tanstack/react-query';
+import { clearSession, getSession, setSession } from '../../utils/actions/session';
+import getAuthUrl from '../../utils/getAuthUrl';
+
+const baseUrl = process.env.NEXT_PUBLIC_API_BASE_URL || '';
+
+export type Role = 'admin' | 'content_manager' | 'member';
+
+export type MiddlewareType = 'auth' | 'guest';
```
edichoska, line 13: lets make this enums, we will have  auto-completion with enums and will reduce the risk of making mistakes.

app/context/authContext.tsx
```diff
@@ -0,0 +1,204 @@
...
+  role: Role;
+};
+
+interface AuthContextType {
+  user?: UserType | null;
+  login: ({
+    // eslint-disable-next-line no-unused-vars
+    email, // eslint-disable-next-line no-unused-vars
+    password, // eslint-disable-next-line no-unused-vars
+    role,
```
edichoska, line 28: can we create an interface here? also for the return type, so if it somehow changes in the future we don't have to remember and make the changes everywhere in the code, we would just change the interface

app/context/authContext.tsx
edichoska, line 13: I am not sure why we have the distinction here for auth and guest. Where would guest be applicable?

edichoska (changes requested):

app/context/authContext.tsx
stefanatanasovmk: Example: on the home page we want to show "Welcome {username}" , for that we would need to get the user object from the `useAuth` hook, but the home page can be viewed as guest as well, so in that case as a `middleware` value in the hook, we would set `guest` so the `useEffect` in the `useAuth` hook don't redirect the guest users to the `/login` route.

components/reusable-components/navigation/Navigation.tsx
```diff
@@ -24,28 +25,44 @@ const Navigation = () => {
...
+            type="link"
+            buttonClass={['primaryButton', 'navigationButton', 'button']}
+            buttonText="Регистрирај се"
+          />
+          <Button
+            href="/login"
+            type="link"
+            buttonClass={['secondaryButton', 'navigationButton', 'button']}
+            buttonText="Најави се"
+          />
```
edichoska, line 41: @MarijaTrajkoska can you revert the changes made in the navigation(adding the buttons)?
We will use this buttons only for the platform. So we will add them again when we are implementing the functionalities regarding the platform

edichoska (approved):
great work here @stefanatanasovmk & @MarijaTrajkoska 
just the small comment and you can merge this

edichoska:
@MarijaTrajkoska @stefanatanasovmk please resolve conflicts and tag me so we can merge this as soon as possible 🙏

edichoska (approved):
great job @stefanatanasovmk @MarijaTrajkoska 
lets test this 🙏

PR #431, opened 2024-08-05
Replace fetch to use custom React Query implementation with AxiosProvider

petark7:
Implement AxiosProvider instead of fetch and replace in every place where normal fetch was used to use React Query instead.

- provide central location in /api for fetch-related files


api/AxiosProvider.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const useAxios = () => {
+  const axiosInstance = useContext(AxiosContext);
+  if (!axiosInstance) {
+    throw new Error('useAxios must be used within an AxiosProvider');
+  }
+  return axiosInstance;
+};
+
+export const AxiosProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
```
edichoska, line 17: refactor to not use React.FC

edichoska: Also resolve conflicts

petark7: Why not use React.FC? I see it being used all over the codebase. For example, it is used in themeContext as well.

petark7: Just resolved the conflicts. The **E2E tests / run\_tests** failed because I have put a check for the presence of ENV variable **NEXT\_PUBLIC\_API\_BASE\_URL** before setting the endpoints. This is not an issue if project is properly setup.

edichoska: @petark7 what is not properly setup?

petark7: @edichoska 

The check I placed in the code checks if the ENV variable '**NEXT\_PUBLIC\_API\_BASE\_URL' exists on the environment variables file, if not, it throws an error. This is what happened during the tests ran in this PR. On my end, there is no issue at all (since I have the ENV var set).**

edichoska (approved):

api/AxiosProvider.tsx
edichoska: ahaaa e2e tests are failing.

edichoska (approved):

PR #433, opened 2024-08-06
Implement tags search functionality

petark7:
This PR implements optimized tags search functionality.

### DIFF
--- a/app/content-panel/tags/page.tsx
+++ b/app/content-panel/tags/page.tsx
@@ -15,6 +15,7 @@ interface Tag {
   id: string;
   name: string;
 }
+
 const Tags = () => {
   const [showAddTag, setShowAddTag] = useState(false);
   const [tags, setTags] = useState<Tag[]>([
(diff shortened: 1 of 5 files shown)


edichoska:
Screenshot 2024-08-09 at 00 39 38
Screenshot 2024-08-09 at 00 40 14
If I add space after the word the search is not working.
The columns in the table shouldnt change their height based on how many entries we have.
If on search there is one item, that row takes the whole height of the table. If we have two items, they split the table height in two.
This shouldn't happen.
You should define the height of the rows to prevent this behaviour.

petark7:
>  If I add space after the word the search is not working. 

Why would we want to still show the tag if we add space after? From what I see, this seems kind of unnecessary since I see no use case for it. 

I have fixed the issue with the height of the rows.

petark7:
> > > If I add space after the word the search is not working.
> > 
> > @edichoska   
> > Why would we want to still show the tag if we add space after? From what I see, this seems kind of unnecessary since I see no use case for it.   
> > I have fixed the issue with the height of the rows.
> 
> Thank you for fixing the row issue. It is better from user experience perspective. Since not only developers will be using this platform, but the marketing team will also be using it. We should always try to remove effort/thinking from the user. We should opt for implementing the following: -a fuzzy search or partial matching algorithm that can handle slight variations or typos in the input
> 
> *   trim leading or trailing spaces or ".,  
>     or we can opt for giving feedback to the user. If the user misspells something we can give suggestions: similar to google when you misspell something while you search

Fixed!

edichoska (approved):

Issue #434, opened 2024-08-12
[FE] Update Login


edichoska:
During login we should implement the following:
- comment out this validation rule since the test password we are using for the content-manager doesn't match the design:   .matches(
        /^(?=.*\d)(?=.*[@$!%*#?&])[A-Za-z\d@$!%*#?&]*$/,
        'Мора да содржи еден број и еден посебен знак'
      ),
we will return it in the future
- if login is successful and the role of the logged user is content-manager we should redirect the user to the content-panel
(you should also handle scenarios where the role is admin, user or if the login fails- credentials dont exist, check API documentation)
- update the route we are using in the submit function to match the one defined in: https://api.learnhub.mk/docs/#authentication-POSTcontent-login

PR #436, opened 2024-08-15
#434 update login

MarijaTrajkoska:


edichoska (commented):
@MarijaTrajkoska @Jasminamih  In content-panel we should have login page, since we need to be calling /content/login
(reuse the current login form component. Here you should make sure we can send the submitLoginForm(formValues) as a prop. So in different pages we can add different functions to be called on login. Example in admin-panel/login etc)

components/reusable-components/login-form/SubmitLoginForm.tsx
```diff
@@ -7,13 +7,13 @@ export interface LoginFormData {
 type LoginFormResponse = any;
 
 export const submitLoginForm = async (formValues: LoginFormData): Promise<LoginFormResponse> => {
-  const url = process.env.NEXT_PUBLIC_API_BASE_URL + '/login'!;
+  const url = `${process.env.NEXT_PUBLIC_API_BASE_URL}/content-login`;
```
edichoska, line 10: check the correct url here: https://api.learnhub.mk/docs/#authentication-POSTcontent-login

components/reusable-components/login-form/LoginForm.tsx
```diff
@@ -51,11 +52,21 @@ const LoginForm = () => {
...
+
+        const userRole = response.role;
+
+        if (userRole === 'content-manager') {
+          window.location.href = '/content-panel';
+        } else if (userRole === 'admin') {
+          window.location.href = '/admin-panel';
+        } else if (userRole === 'user') {
+          window.location.href = '/user-dashboard';
+        }
```
edichoska, line 65: Since we are making a call to a route that is intended for login of content manager, we have no use of this logic here.

edichoska:
@MarijaTrajkoska @Jasminamih This should have been a draft PR

edichoska (changes requested):

components/reusable-components/login-form/SubmitLoginForm.tsx
edichoska: also you should use the login from authContext

components/reusable-components/login-form/LoginForm.tsx
edichoska: also you have a similar function in utils, **getAuthUrl**

components/reusable-components/login-form/LoginForm.tsx
```diff
@@ -35,36 +33,31 @@ const LoginForm = ({ submitLoginForm }: LoginFormProps) => {
...
-
     try {
-      const response = await submitLoginForm(formValues);
-      if (response) {
-        toast.success(response);
-
+      const response = await login({
+        email: values.email,
+        password: values.password,
+        role: Role.content_manager,
```
edichoska, line 46: @Jasminamih Have you tested this?
As you can see in the API documentation(https://api.learnhub.mk/docs/#authentication-POSTcontent-login), the route we should call is `https://staging-api.learnhub.mk/content/login`. But in the `Role` enum we have defined content manager as: `content_manager = 'content_manager',` . Because of this the route you will call is `https://staging-api.learnhub.mk/content-manager/login`.  That route doesn't exist.

edichoska (approved):

components/reusable-components/login-form/LoginForm.tsx
edichoska, line 1: LoginForm should stay reusable. We shouldn't place the validation and the submit routes/logic there

edichoska (changes requested):

Types/index.ts
edichoska, line 1: types folder should have the same naming convention as the other folders in our project

edichoska: You should rename the index.ts folder to something that will give an indication of the content of the types inside.
If we are going to be extracting the types in separate files, we need a structure that will be easy to follow.

app/context/authContext.tsx
edichoska, line 1: @mitkapanarin why is the setting the session removed from this file?
When the user logs in we should create the session

app/signup/page.tsx
edichoska, line 1: Signup logic will be only used for members. Lets remove this file for now.

utils/getAuthUrl.ts
```diff
@@ -3,7 +3,7 @@ import { Role } from '../app/context/authContext';
 export default function getAuthUrl(baseUrl: string, role: Role): string {
   let url = baseUrl;
   if (role === 'member') url = `${baseUrl}`;
-  else if (role === 'content_manager') url = `${baseUrl}/content`;
+  else if (role === 'content_manager' || role === 'content') url = `${baseUrl}/content`;
```
edichoska, line 6: we don't have role content

edichoska (changes requested):

Types/index.ts
mitkapanarin: I changed the name of the folder because I had some eslint issues while I was commiting code, that I think were conected to the folder name

PR #439, opened 2024-08-21
Tags page with React Query fetching implementation

petark7:
The PR implements the API and allows adding/editing, searching and deleting tags.

Currently, the token is manually added in the ENV file but as soon as the auth implementation is ready, that is to be replaced.

The page is almost complete, only thing missing is a pagination component allowing to switch to different page within the table data.


app/content-panel/tags/page.tsx
```diff
@@ -1,15 +1,153 @@
...
+      return { success: false, error: 'API Error' };
+    }
+  };
+
+  const handleSaveChanges = (tagId: string, newName: string) => {
+    setTags((prevTags) =>
+      prevTags.map((tag) => (tag.id === tagId ? { ...tag, name: newName.trim() } : tag))
+    );
+    setEditingTagId(null);
+  };
```
edichoska, line 82: Lets refetch the data so we ensure that the data is always in sync with the server, and we won't have stale or inconsistent data. Here the dataset is relatively small so we should opt for refetching.

edichoska: in all crud operations here

petark7: Done!

app/content-panel/tags/page.tsx
```diff
@@ -1,15 +1,157 @@
...
+
+    try {
+      await addNewTagMutation.mutateAsync({
+        tagName: trimmedNewTag,
+      });
+      await refetch();
+
+      // If the mutation was successful, add the new tag to the local state
+      const newId = tags.length > 0 ? (parseInt(tags[tags.length - 1].id, 10) + 1).toString() : '1';
+      setTags([...tags, { id: newId, name: trimmedNewTag }]);
```
edichoska, line 71: since we implemented refetch you no longer have need to manipulate setTags like this

app/content-panel/tags/page.tsx
```diff
@@ -1,15 +1,157 @@
...
+    }
+  };
+
+  const handleSaveChanges = async (tagId: string, newName: string) => {
+    try {
+      await editTagMutation.mutateAsync({ tagId, newName: newName.trim() });
+      await refetch();
+      setTags((prevTags) =>
+        prevTags.map((tag) => (tag.id === tagId ? { ...tag, name: newName.trim() } : tag))
+      );
```
edichoska, line 85: same here, and all the places in the PR where you have refetch

edichoska (approved):
This looks amazing, kudos for the clean code 👏

PR #449, opened 2024-11-11
Implement login authentication using NextAuth

AnetaDimoska:
Changes implemented:

- Integrated NextAuth with the credentials provider to handle user login via email and password.

- Implemented session management using JWT strategy with custom callbacks to store accessToken, userId, and other user information.

- Created custom authentication options in authOptions.ts to connect with our backend API for user verification.

- Extended NextAuth types to include custom fields like accessToken and userId.
- Updated the next-auth.d.ts file to handle custom session data and resolved TypeScript errors.
- Secured the content-panel/dashboard route using server-side checks with getServerSession. If a user is not authenticated, they are redirected to the /login page. (for testing purposes)
- Added server-side protection to ensure only authenticated users can access the dashboard.  (for testing purposes)

How to Test:

- Navigate to content-panel/login and attempt to log in using valid content panel credentials.

- Verify that authenticated users are redirected to content-panel/dashboard and can access protected content.
- Test the logout functionality to ensure it clears the session and redirects users back to  content-panel/login.


app/content-panel/dashboard/page.tsx
```diff
@@ -1,7 +1,25 @@
...
-  return <div>content dashboard</div>;
+import { authOptions } from '../../../utils/authOptions';
+import LogoutButton from '../../../components/reusable-components/button/LogoutButton';
+
+const Dashboard = async () => {
+  const session = await getServerSession(authOptions);
+
+  if (!session) {
+    redirect('/content-panel/login');
+  }
```
edichoska, line 13: Since we will handle authorisation in a middleware, this won't be needed

app/context/authContext.tsx
edichoska, line 1: Remove the commented out code. If we need the code in the future we can create a branch from a certain commit, that is the beauty of using version control.

components/reusable-components/login-form/LoginForm.module.scss
```diff
@@ -149,3 +155,16 @@
 .checkboxDark {
   border: 1.5px solid var(--Secondary, hsla(220, 20%, 43%, 1));
 }
+.errorMessageContainer {
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  background-color: #fff1f1; /* Light red background for emphasis */
+  border: 1px solid #ff6b6b; /* Slightly darker border */
```
edichoska, line 163: we have predifined colors and styles you can see them in _colors.scss and typography.scss, please incorporate  those styles in the new classes you have added here

components/reusable-components/login-form/LoginForm.tsx
```diff
@@ -81,40 +87,22 @@ const LoginForm: React.FC<LoginFormProps> = ({
...
-            <Image className={styles.socialIcon} src={linkedin} alt="Linkedin" />
-          </Link>
-        </div>
-        <p>
-          Немаш креиран профил?{' '}
-          <Link href="/signup" className={styles.signup}>
-            Регистрирај се
-          </Link>
-        </p>
-      </div>
```
edichoska, line 117: why is this part of the component deleted? This is the reusable login component that will be used all over the site: admin-panel, content-panel and platform. And it is done as per the Figma Design

components/reusable-components/login-form/LoginFormOld.tsx
edichoska, line 1: Remove all files now renamed Old.
We have no need to hoard components, we can always go back if something breaks/doesnt work

utils/authOptions.ts
edichoska, line 1: Clean this file up. We shouldn't leave commented out code. It would be challenging for whomever works in this file next to know if he/she should uncomment this, will it break something etc

app/context/authContext.tsx
```diff
@@ -24,26 +27,55 @@ export const AuthProvider: React.FC<{ children: ReactNode }> = ({ children }) =>
...
-      // eslint-disable-next-line camelcase
-      saveToLocalStorage('session', { token: access_token, role: user.role });
+      if (!result?.ok) {
+        throw new Error(result?.error || 'Failed to login');
+      }
+      return result;
+    },
+    onSuccess: () => {
+      // Redirect to the dashboard on success
+      router.push('/content-panel/dashboard');
```
edichoska, line 50: can we make this a parametar that will pass? this way the login could be reused for admin-panel as well?

utils/authOptions.ts
```diff
@@ -0,0 +1,120 @@
...
+      const newSession = { ...session };
+      newSession.accessToken = token.accessToken;
+      newSession.user.id = token.id;
+      newSession.user.email = token.email;
+      newSession.user.name = token.name;
+      return newSession;
+    },
+  },
+  pages: {
+    signIn: '/control-panel/login', // Custom login page path
```
edichoska, line 117: how can we expand this to work with the admin-panel and platform as well?

edichoska (changes requested):
Overall great work @AnetaDimoska 👏

edichoska:
@AnetaDimoska Please resolve conflicts as well 🙏

app/context/authContext.tsx
AnetaDimoska: Yes, i made changes and this url is now dynamic.

components/reusable-components/login-form/LoginForm.tsx
AnetaDimoska: I have reworked the LoginForm.tsx and now it accepts only one prop and that is onSubmit function.
I changed the previous one because i found it anti-next.js pattern to make the login page.tsx file a client component.
I prefer ( and also as a best practice ) to keep the page.tsx files server components.

Since, no hooks are allowed in a server component, i created two files that serves as containers where i import the necessary hooks and send prop values to LoginForm.tsx ( path: components/reusable-components/login-form ). Then, i import those components in the page.tsx files (one for content-panel and another for admin-panel), therefore keeping the file server component.

utils/authOptions.ts
AnetaDimoska: I have removed this for now, since it is not mandatory for configuration.
This is mainly used to customize the default authentication pages provided by NextAuth.

As far as i have researched, it is possible to make this reusable too, but i suggest to get back to this once we have finished with the middleware and the whole auth process.

AnetaDimoska:
@edichoska I can't see the option for Resolving Conflicts, it is disabled. Can you please assist me with this? 🙏

Tikkurilla (changes requested):

api/authApi.ts
```diff
@@ -0,0 +1,123 @@
+import { LoginParams, LoginResponse, Role, UserType } from '../Types';
+import {
+  getFromLocalStorage,
+  removeFromLocalStorage,
+  saveToLocalStorage,
+} from './utils/localStorageUtils';
+
+const baseUrl = process.env.NEXT_PUBLIC_API_BASE_URL || '';
+
+export const getUser = async (): Promise<UserType | null> => {
```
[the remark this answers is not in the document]
AnetaDimoska: Thank you for the feedback! I agree that implementing a consistent query approach and proper error handling is essential for the project. However, I believe these changes would require more time and effort as they impact few parts of the code:

**Standardizing Axios queries**: This involves aligning the data fetching logic across different functions, which would require thorough updates and testing to ensure nothing breaks.

**Error handling**: Designing and implementing robust error handling (e.g., appropriate error messages, UI feedback, etc.) would also take time and need to be applied consistently across the project.

Considering this, I suggest merging my current PR so that other developer can continue working on the middleware without blocking progress. Once this is merged, we can create two new tasks to:

- Synchronize the data fetching logic using the axios approach.

- Develop and implement a consistent error-handling strategy.

This way, we can make these improvements without delaying ongoing work.
Let me know what you think on this.

edichoska (approved):

api/authApi.ts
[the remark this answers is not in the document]
Tikkurilla: yep, makes sense :)

Tikkurilla (approved):

PR #450, opened 2024-11-15
[FE] Content-Panel: Add New Blog Article

petark7:
This PR fully implements creating a blog article.

*   [x] Fully conforms with the backend
*   [x] Is tested (creating blog posts, adding tags and creating tags) and is working
*   [x] Implements form validation via Formik
*   [x] It allows selecting and creating tags on the go™
*   [x] The page component is server component and the rest of the code is broken up in smaller, client-components.

Currently, the created blog post is immediately sent for review. When the backend has option for drafts, that can be additionally implemented.

**Note: To test the feature, you need to be logged in as content-manager and have the token set in the .env file.**


apis/mutations/blogs/useAddNewPost.ts
```diff
@@ -1,23 +1,40 @@
...
 import ENDPOINTS from '../../endpoints';
+import QUERY_KEYS from '../../queryKeys';
 
 type NewPost = {
   title: string;
-  body: string;
-  userId: number;
+  excerpt: string;
+  content: string;
+  tags: number[] | string[];
```
edichoska, line 14: I would assume tags should be only string[]. What is the usecase for number[]?

components/module-components/blog/TagInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const addNewTag = async (tagName: string) => {
+    const newTag = await addNewTagMutation.mutateAsync({ tagName });
+    onTagsChange([...selectedTags, newTag.data as TagObject]);
+
+    setSearchTag('');
+    setFilteredTags([]);
+  };
+
+  useEffect(() => {
+    if (data?.data) {
```
edichoska, line 41: We should avoid coupling between the API layer and our components. We shouldn't transform the response data inside the components. (data.data)
This should be done inside the useGetTags query.

edichoska (approved):

apis/mutations/blogs/useAddNewPost.ts
petark7: Apparently, no usecase.. for some reason I thought since user can send either list of integers or a list of strings, I should enable that. However, thinking about it now, there is no need for this. Will modify.

components/module-components/blog/TagInput.tsx
petark7: Thank you for your feedback!

I want to make sure I fully understand your suggestion. Are you recommending that:

*   The useGetTags hook should return only the data field (omitting fields like "links" and "meta") to simplify the response? (this kind of seems unnecessary to me... what if we need to use the other fields somewhere else?)
*   Or, that the transformation logic (e.g., converting name to lowercase) should be moved to the useGetTags hook itself, so the component receives already-processed data?

Let me know so I can better address the issue.

edichoska: @petark7 
I’m recommending the second option: that transformation logic (e.g., processing data.data, converting names to lowercase) should be moved into the useGetTags hook. This will make sure that the component receives data that’s already in the desired format, we will have better separation of concerns and we will reduce the chance of repetitive logic in different components.

To clarify, I’m not suggesting the hook should not return fields like links or meta, as they might be useful elsewhere.
(you can use the select option in useQuery to return all of them).

The goal is just to avoid components having to directly manipulate the raw API response (like data.data), which keeps them simpler and focused on rendering.

edichoska: Since this is just a suggestion, I am also approving the PR.

edichoska (dismissed):
⭐

components/module-components/blog/TagInput.tsx
edichoska: Since this is just a suggestion, I am also approving the PR.

edichoska (approved):
This looks good now. 🚀

PR #460, opened 2024-12-02
[FE] Content-Panel: Blogs: General Improvements

petark7:
## Summary of Changes

*   adjusted the design in blogs page to match the tags page
*   added skeletons while fetching data and improved the message UI if no data is available
*   translated all fields (both from Tags and Blogs) to Macedonian
*   improved table UI design
*   removed unnecessary /dashboard route

## Manual Test Steps

1.  Go to /content-panel/tags or /content-panel/blogs and observe loading state and "no data" message. 
2.  Observe the similarity in the designs between those two routes (tags - blogs)

Both pages should be mobile responsive.

# Screenshots


components/module-components/create-blogs/BlogManagementControls.tsx
edichoska, line 1: Can you create one component for search that would be reused for tags and blogs?

components/reusable-components/reusable-table/reusableTable.module.scss
```diff
@@ -41,14 +49,39 @@ $table-height: 400px;
...
+  }
+}
+
+.skeletonCell {
+  background-color: #e0e0e0; /* Light gray for skeleton */
+  border-radius: 4px;
+  height: 20px; /* Skeleton height */
+  width: 100%; /* Fills the cell width */
+  animation: skeleton-loading 1.5s infinite;
+}
```
edichoska, line 75: Remove comments

edichoska (changes requested):

components/module-components/create-blogs/BlogManagementControls.tsx
petark7: Can you elaborate how you have imagined this component? Because as it is, I am already reusing the \<Input/> component in both places and just change how the state is handled and the onChange functionality.

edichoska: We already have a Search component. It would be nice if that component works, and we apply these styles to it. So we can reuse it in the admin-panel and platform. Mostly so we have consistency in the UI components

PR #466, opened 2024-12-09
[FE] Implement Role-Based Middleware for Page Authorization [444]

petark7:
### Changes
- Migrated from custom localStorage-based authentication to NextAuth
- Removed unnecessary authentication context (`authContext.tsx`)
- Deleted obsolete authentication API files (`authApi.ts`, `session.ts`)
- Updated axios instance to use NextAuth session for token management
- Created new login and logout hooks in queries folder
- Added basic unauthorized page
- Implemented middleware to protect routes

### Key Improvements
- More secure authentication method
- Simplified authentication flow
- Better integration with Next.js authentication patterns
- Removed client-side token management

### Breaking Changes
- Removed custom authentication context
- Changed login and logout mechanisms
- Requires updates to environment variables (added `AUTH_SECRET`, `NEXTAUTH_URL`)

### Testing instructions
-- Go to route /content-panel/tags (a protected route)
- you should not be allowed and redirected to content-panel/login

-- Login using content-manager account at /content-panel/login
- try accessing the protected route /content-panel/tags and modify tag. It should be successful.
- try accessing /admin-panel/  -> you will land on unauthorized page


apis/mutations/login/useLogin.tsx
```diff
@@ -0,0 +1,57 @@
...
+        cfTurnstileResponse,
+        userType,
+        remember,
+      });
+
+      if (!result?.ok) {
+        throw new Error(result?.error || 'Неуспешна најава');
+      }
+
+      router.push(redirectUrl);
```
edichoska, line 45: It would be better to not reroute inside the mutation. Because of the asynchronous nature of mutations here is not the best place to handle side effects.
We can reroute inside the component , when we use the mutation, based on the mutation state.

edichoska (approved):
great work 🚀

apis/mutations/login/useLogin.tsx
petark7: Good point. Thanks for the feedback!

edichoska (approved):

Issue #367, opened 2024-06-16
[FE] Content-Panel Fix Search and Filter functionality


edichoska:
**Summary**
The search and filter functionality on the content panel blogs and tags pages is not working as expected. Entering search terms or selecting filters does not trigger any changes.

**Steps to Reproduce**
1. Go to the content-panel/blogs or content-panel/tags page
2. In the search bar, type any query (e.g article title).
3. Alternatively, try selecting any available filters (e.g., role, status).
4. Observe the results after applying the search or filter.

**Expected Result**
The list of items should update to display only the relevant results based on the search term or selected filters.

**Actual Result**
No change occurs in the list. The search and filter functionality does not return any results, as if no actions were taken.

PR #467, opened 2024-12-09
#367 content panel fix search and filter functionality

Jasminamih:
## Summary of Changes

• Implemented a dynamic search input for blogs, users can now type in the search input to display matching results.
• Resolved existing TypeScript errors.
• Removed the filter from blogs page.


edichoska (changes requested):
Not sure what happened here, why we have 61 changes?

PR #473, opened 2024-12-16
Publish blog posts 464

aleksandarmicev:
Whit this PR, we made changes to include a draft status, providing different options for content managers or admins based on who is logged in. 
Admins now have the ability to publish blogs. Additionally, the status for API data has been changed to "Draft" because the API was returning an undefined status for fetched data. I believe this issue needs to be addressed by the backend team.


app/content-panel/blogs/create/page.tsx
```diff
@@ -1,15 +1,33 @@
...
 import PublishArticleForm from '../../../../components/module-components/blog/PublishArticleForm';
 
+type UserRole = 'content_manager' | 'admin';
+
 const PostArticle = () => {
+  const { data: session, status } = useSession();
+
+  if (status === 'loading') {
+    return <div>Loading....</div>;
+  }
```
edichoska, line 17: Usually when we are on a page example here blogs, and we show indication of loading it should indicate that the posts are loading not the status of the logged in user.
I think it would be helpfull if you check what we get from useSession

app/content-panel/blogs/create/page.tsx
```diff
@@ -1,15 +1,33 @@
...
+
+  if (!session || !session.user.role) {
+    return <div>Access denied</div>;
+  }
```
edichoska, line 21: Also this here, we have middleware now that handles access.
Is the middleware not performing as expected or you wanted to handle something else here?

components/module-components/blog/PublishArticleForm.tsx
```diff
@@ -10,7 +10,11 @@ import TiptapEditor from '../../editor/TiptapEditor';
 import TagManager from './TagManager';
 import Button from '../../reusable-components/button/Button';
 
-const PublishArticleForm = () => {
+interface PublishArticleFormProps {
+  userRole: 'content_manager' | 'admin';
```
edichoska, line 14: we should already have some enum containing the roles for the user, please reuse that one. In case we don't have it, please create it.

edichoska (changes requested):

edichoska:
@aleksandarmicev Can you please go though the comments above. If anything is already done, click on Resolve Conversation. If the PR is ready to be rechecked please rerequest a review by clicking on the red icon next to my name in Reviews on top right

app/content-panel/blogs/create/page.tsx
```diff
@@ -1,15 +1,25 @@
...
+import React from 'react';
+import { useSession } from 'next-auth/react';
 import styles from './createArticlePage.module.scss';
 import PublishArticleForm from '../../../../components/module-components/blog/PublishArticleForm';
+import { UserRole } from '../../../../Types';
 
 const PostArticle = () => {
+  const { data: session } = useSession();
+
+  // console.log('Session data:', session);
```
edichoska, line 14: remove console.log

components/module-components/blog/PublishArticleForm.tsx
```diff
@@ -28,8 +36,12 @@ const PublishArticleForm = () => {
       .min(1, 'Мора да селектираш барем еден таг.'),
   });
 
-  const handleAddPost = (values: NewPost) => {
-    addNewPostMutation.mutate(values);
```
edichoska, line 32: why is handleAddPost deleted?

app/content-panel/blogs/create/page.tsx
```diff
@@ -1,15 +1,25 @@
...
+  const userRole = session?.user.role as UserRole;
+
   return (
     <div className={styles.container}>
       <h2>Објави статија</h2>
 
       <div className={styles.controlsContainer}>
-        <PublishArticleForm />
+        <PublishArticleForm userRole={userRole} />
```
edichoska, line 22: The main purpose of the create blog page should be to create/add new blog. On create we should call this endpoint: https://api.learnhub.mk/docs/#content-POSTcontent-blog-posts
We can change the status of already created blogs. That is why you need to send the id of the blog.

edichoska (changes requested):

components/module-components/blog/PublishArticleForm.tsx
```diff
@@ -3,16 +3,20 @@
...
 import TiptapEditor from '../../editor/TiptapEditor';
 import TagManager from './TagManager';
 import Button from '../../reusable-components/button/Button';
+import { UserRole } from '../../../Types';
 
 const PublishArticleForm = () => {
   const addNewPostMutation = useAddNewPost();
   const [selectedTags, setSelectedTags] = useState<TagObject[]>([]);
+  const { data: session } = useSession();
+  const userRole = session?.user.role as UserRole;
```
edichoska, line 19: we should avoid type assertions

components/module-components/blog/PublishArticleForm.tsx
```diff
@@ -104,6 +109,18 @@ const PublishArticleForm = () => {
...
           </div>
 
+          <div className={styles.fields}>
+            <label htmlFor="status" className={styles.inputLabel}>
+              Статус
+            </label>
+            <Field as="select" name="status" className={styles.dropdown}>
+              <option value="draft">Draft</option>
+              <option value="in_review">In Review</option>
+              {userRole === UserRole.admin && <option value="published">Published</option>}
```
edichoska, line 119: declare these options in an array so we can easily change them if needed in the future, and we wouldn't need to "hunt" them in the code

components/module-components/blogs/BlogListView.tsx
```diff
@@ -35,15 +39,37 @@ const BlogListView = () => {
     router.push(`/content-panel/blogs/${id}`);
   };
 
+  const handleStatusChange = (id: string, status: string) => {
+    updateStatus({ id, status }); // Користење на мутацијата за да го ажурирате статусот
```
edichoska, line 43: i don't think we need this comment here, the function is pretty simple and straight forward

apis/mutations/blogs/useAddNewPost.ts
```diff
@@ -34,6 +37,7 @@ const useAddNewPost = () => {
     onSuccess: () => {
       queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
       toast.success('Статијата беше успешно објавена!');
+      router.push('/content-panel/blogs');
```
edichoska, line 40: It would be better to not reroute inside the mutation. Because of the asynchronous nature of mutations here is not the best place to handle side effects.
We can reroute inside the component , when we use the mutation, based on the mutation state.

edichoska (changes requested):

edichoska (changes requested):
Please update this branch with latest staging so we can properly test this functionality.

edichoska (commented):
great work @aleksandarmicev 
Just a couple of small chages:
1. In the table display the status options capitalized. Don't use snake_case
2. Add the functionality to change the status inside the view and edit. The logged in user should be able to see the status of the blog post while in view mode. And it would be better if the user can change the status while he is editing the post.

components/module-components/blog/StatusManager.tsx
```diff
@@ -0,0 +1,46 @@
...
+import styles from './PublishArticleForm.module.scss';
+
+export const capitalizeFirstLetter = (string: string) => {
+  return string.charAt(0).toUpperCase() + string.slice(1).replace(/_/g, ' ');
+};
+
+export const statusOptions = [
+  { value: 'draft', label: 'Draft' },
+  { value: 'in_review', label: 'In Review' },
+];
```
edichoska, line 13: Just a suggestion: Here a good approach would be to use `const type assertion`. We will treat the object as deeply immutable and will provide read-only access.  We will also include all options before hand, instead of mutating the object based on the user role.
Then you can conditionally render the status based on the role of the logged in user.

components/reusable-components/blogDetails-card/BlogDetailsCard.tsx
```diff
@@ -29,8 +33,11 @@ const BlogDetailsCard: React.FC<BlogDetailsCardProps> = ({
   const [showModal, setShowModal] = useState(false);
   const [modalType, setModalType] = useState<'back' | 'cancel'>('back');
   const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
+  const [currentStatus, setCurrentStatus] = useState(status);
+  const [tempStatus, setTempStatus] = useState(status);
```
edichoska, line 37: If the current status is the status of the blog post before editing do you think there is a better approach to save/show that value instead of using two useStates?

edichoska (changes requested):

api/utils/blogStatusUtils.ts
edichoska, line 1: Here a better naming would be:  `capitalizeAndFormatString`. Since we can use this function anywhere in the project where we need to format a string, it is not tied just to the blogs feature.
(the name is a suggestion for a direction, you may think of a different one)

edichoska (changes requested):
🤞

edichoska (approved):
amazing! keep up the great work @aleksandarmicev 💪

PR #474, opened 2024-12-19
Fix search input functionality 367

Jasminamih:
## Summary of Changes
• Fixed the search input for blogs. Users can now type in the search input to display matching results.


edichoska:
thanks for the fix @Jasminamih  🚀  Now it works as it should

edichoska (approved):

PR #477, opened 2025-01-27
Bug: View Blog Post 476

aleksandarmicev:
With these changes, we're resolving an error that showed up when we clicked on a single blog post.

 The first error was in getBlogs.ts, where the 'status' parameter had the value "published," and all the blog posts had that status.
Additionally, there was an incorrect path for fetching blog posts, and by correcting the path, the data now gets the value from the API. Finally, when displaying a single blog post, there was a problem showing the value for the Author and tags, which was resolved by correcting BlogDetailsCard.tsx. The constant value in “renderInput” is now replaced with the “blogAuthor” value from the API. For tags, we corrected the mapping of tag names to properly show tag values.

### DIFF
--- a/apis/queries/blogs/getBlogDetails.ts
+++ b/apis/queries/blogs/getBlogDetails.ts
@@ -2,12 +2,11 @@ import { useQuery } from '@tanstack/react-query';
 import { getSession } from 'next-auth/react';
 import QUERY_KEYS from '../../queryKeys';
 import { BlogDetailsData } from '../../../components/reusable-components/_Types';
+import ENDPOINTS from '../../endpoints';
 
 const fetchBlogDetails = async (id: string): Promise<BlogDetailsData> => {
-  const url = `${process.env.NEXT_PUBLIC_API_BASE_URL}/content/blog-posts/${id}`;
-
   const session = await getSession();
-  const response = await fetch(url, {
+  const response = await fetch(`${ENDPOINTS.BLOGS.GET_ALL}/${id}`, {
     method: 'GET',
     headers: {
       'Content-Type': 'application/json',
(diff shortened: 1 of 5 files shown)


apis/queries/blogs/getBlogDetails.ts
```diff
@@ -1,24 +1,32 @@
...
-    image: blogPost.image || '',
-    content: blogPost.content || 'N/A',
+    title: blogPost.title,
+    image: blogPost.image,
+    content: blogPost.content,
     author: {
-      firstName: blogPost.author?.firstName || blogPost.author?.first_name || 'N/A',
-      lastName: blogPost.author?.lastName || blogPost.author?.last_name || 'N/A',
+      first_name: blogPost.author.first_name,
+      last_name: data.data.author.last_name,
```
edichoska, line 29: since you have declared   const blogPost = data.data you can use it here like you did for first_name

apis/queries/blogs/getBlogDetails.ts
```diff
@@ -1,24 +1,32 @@
 import { useQuery } from '@tanstack/react-query';
+import { getSession } from 'next-auth/react';
 import QUERY_KEYS from '../../queryKeys';
 import { BlogDetailsData } from '../../../components/reusable-components/_Types';
 
 const fetchBlogDetails = async (id: string): Promise<BlogDetailsData> => {
-  const url = `${process.env.NEXT_PUBLIC_API_BASE_URL}/blog-posts/${id}`;
-  const response = await fetch(url);
+  const url = `${process.env.NEXT_PUBLIC_API_BASE_URL}/content/blog-posts/${id}`;
```
edichoska, line 7: can you declare the url in endpoints?

edichoska (changes requested):
on some blog posts i am not getting the date, can you investigate further of what is happening. other than that, the bug seems to be resolved.

edichoska (approved):
⭐

PR #482, opened 2025-02-23
fix (invisible actions menu): actions menu is correctly displayed

gogostanoev:
The actions menu on the last blog post on the list is no longer invisible, promting the user to scroll a bit more down. Instead it faces upward.


edichoska (changes requested):
Screenshot 2025-02-23 at 20 51 33
Since this is happening with the next to last Actions as well, it might be better to make the Actions appear above the table. I would also revert the change where the last one appears on top of Actions Button.

edichoska (approved):
💅

Issue #490, opened 2025-03-12
[FE] FAQ Section Improvement


aleksandarmicev:
FAQ section improvement

PR #491, opened 2025-03-12
Improve FAQ section #490

aleksandarmicev:
Improvement in FAQ section: 
- handleClickedTopic function is changed from div to li element to increase the space for opening the answers.
- reducing repetition for theme with creating different const


dichoskaemilija:
@aleksandarmicev you need to create another PR which will be merged to main. For this you need to create a branch from main, and when you are creating the PR you need to choose the base: main. This way we won't move all the staging changes to main.

dichoskaemilija (approved):
⭐ x2

PR #492, opened 2025-03-13
Improve FAQ section-main

aleksandarmicev:
Improvement in FAQ section:

- handleClickedTopic function is changed from div to li element to increase the space for opening the answers.
- reducing repetition for theme with creating different const


dichoskaemilija (approved):
⭐

PR #506, opened 2025-04-01
494 fe content panel bug tags

### LINKED ISSUE #494: [FE] Content-Panel: Bug: Tags
Regardless if the user is in edit mode or not, tags can be added. 
We need to check the role of the logged in user, because only admins should be able to create new tags.

mitkapanarin:
Tag Management Requirements:

Both content managers and admins can add existing tags
Only admins can create new tags
All operations require edit mode

### DIFF
--- a/apis/mutations/blogs/useAddNewPost.ts
+++ b/apis/mutations/blogs/useAddNewPost.ts
@@ -20,7 +20,7 @@ type ErrorResponse = {
   statusCode?: number;
 };
 
-const useAddNewPost = () => {
+const useAddNewPost = (onSuccess?: () => void) => {
   const queryClient = useQueryClient();
   const axios = useAxios();
 
@@ -33,8 +33,12 @@ const useAddNewPost = () => {
       toast.error(error?.response?.data?.message || 'Настана грешка при креирање на статијата.');
     },
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
       toast.success('Статијата беше успешно објавена!');
+      onSuccess?.();
+      // Invalidate queries after showing success message
+      setTimeout(() => {
+        queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
+      }, 100);
     },
   });
 };
--- a/apis/mutations/blogs/useAddNewPost.ts
+++ b/apis/mutations/blogs/useAddNewPost.ts
@@ -40,5 +40,3 @@ export const useAddNewPost = (onSuccess?: (data: any, post: NewPost) => void) =>
     },
   });
 };
-
-export default useAddNewPost;
--- a/apis/mutations/blogs/useAddNewPost.ts
+++ b/apis/mutations/blogs/useAddNewPost.ts
@@ -20,25 +20,23 @@ type ErrorResponse = {
   statusCode?: number;
 };
 
-const useAddNewPost = (onSuccess?: () => void) => {
+export const useAddNewPost = (onSuccess?: (data: any, post: NewPost) => void) => {
   const queryClient = useQueryClient();
   const axios = useAxios();
 
   return useMutation({
-    mutationFn: async (newPost: NewPost) => {
-      const response = await axios.post(ENDPOINTS.BLOGS.CREATE, newPost);
-      return response.data;
+    mutationFn: (newPost: NewPost) =>
+      axios.post(ENDPOINTS.BLOGS.CREATE, newPost).then((response) => response.data),
+
+    onSuccess: (data, variables) => {
+      toast.success('Статијата беше успешно објавена!');
+      queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
+
+      onSuccess?.(data, variables);
     },
+
     onError: (error: AxiosError<ErrorResponse>) => {
       toast.error(error?.response?.data?.message || 'Настана грешка при креирање на статијата.');
-    },
-    onSuccess: () => {
-      toast.success('Статијата беше успешно објавена!');
-      onSuccess?.();
-      // Invalidate queries after showing success message
-      setTimeout(() => {
-        queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
-      }, 100);
     },
   });
 };
--- a/apis/mutations/blogs/useEditBlogPost.ts
+++ b/apis/mutations/blogs/useEditBlogPost.ts
@@ -20,21 +20,22 @@ type ErrorResponse = {
   statusCode?: number;
 };
 
-const useEditBlogPost = () => {
+const useEditBlogPost = (onSuccess?: () => void) => {
   const queryClient = useQueryClient();
   const axios = useAxios();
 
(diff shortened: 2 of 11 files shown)


apis/mutations/blogs/useAddNewPost.ts
```diff
@@ -33,8 +33,12 @@ const useAddNewPost = () => {
       toast.error(error?.response?.data?.message || 'Настана грешка при креирање на статијата.');
     },
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
       toast.success('Статијата беше успешно објавена!');
+      onSuccess?.();
+      // Invalidate queries after showing success message
+      setTimeout(() => {
+        queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
+      }, 100);
```
dichoskaemilija: You dont need to call setTimeout inside onSucess. Follow the pattern bellow:
export const useAddNewPost = () => {
  const queryClient = useQueryClient();
  return useMutation((params: Params) => addNewPost(params), {
    onSuccess: (_, variables) => {
//Here the query key should be the one from the get blog posts query
      queryClient.invalidateQueries({ queryKey: ["newPost"] });
    },
  });

apis/mutations/blogs/useEditBlogPost.ts
```diff
@@ -45,8 +46,12 @@ const useEditBlogPost = () => {
       toast.error(error?.response?.data?.message || 'Настана грешка при ажурирање на статијата.');
     },
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
       toast.success('Статијата беше успешно ажурирана!');
+      onSuccess?.();
+      // Invalidate queries after showing success message
+      setTimeout(() => {
+        queryClient.invalidateQueries({ queryKey: QUERY_KEYS.BLOGS.ALL });
+      }, 100);
```
dichoskaemilija: same from above

apis/mutations/blogs/useEditBlogPost.ts
```diff
@@ -20,21 +20,22 @@ type ErrorResponse = {
...
+      const [editResponse, statusResponse] = await Promise.all([
+        axios.patch(ENDPOINTS.BLOGS.EDIT(id), {
+          title,
+          content,
+          tags,
+        }),
+        axios.patch(`${ENDPOINTS.BLOGS.UPDATE_STATUS(id)}`, {
+          status,
+        }),
+      ]);
```
dichoskaemilija: what happens id the user doesn't update the status? we shouldnt make the second call in that scenario

apis/mutations/blogs/useEditBlogPost.ts
```diff
@@ -20,21 +20,22 @@ type ErrorResponse = {
   statusCode?: number;
 };
 
-const useEditBlogPost = () => {
+const useEditBlogPost = (onSuccess?: () => void) => {
```
dichoskaemilija: Is there a reason we are accepting onSuccess as a parametar here?
the useEditBlogPost handles the onSuccess inside it. It returns isSuccess value.

components/module-components/blog/PublishArticleForm.tsx
```diff
@@ -1,21 +1,32 @@
...
+  const isAdmin = session?.user?.role === UserRole.admin;
   const router = useRouter();
+  const tagManagerRef = useRef<TagManagerRef>(null);
+  const [selectedTags, setSelectedTags] = useState<Tag[]>([]);
+  const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
+
+  const addNewPostMutation = useAddNewPost(() => {
+    setHasUnsavedChanges(false);
+    router.push('/content-panel/blogs');
+  });
```
dichoskaemilija: Here you can use the following pattern:
call the addNewPostMutation mutation, on Success handle the    
setHasUnsavedChanges(false);
router.push('/content-panel/blogs');

on Error display a toast with the error

components/module-components/blog/TagInput.tsx
```diff
@@ -12,20 +12,38 @@ import { Tag } from '../../reusable-components/_Types';
...
+  const { selectedTags, onTagsChange, isAdmin = false, isEditMode = false } = props;
   const [searchTag, setSearchTag] = useState<string>('');
   const debouncedSearchTerm = useDebounce(searchTag, 300);
   const { data } = useGetTags(debouncedSearchTerm);
   const addNewTagMutation = useAddNewTag();
 
+  useImperativeHandle(ref, () => ({
+    clearInput: () => setSearchTag(''),
+  }));
+
```
dichoskaemilija: What is the purpose of useImperativeHandle?

dichoskaemilija (changes requested):
Great start! Just a few improvements suggested.
Can you please go through all the changed files and remove the comments 🙏

components/module-components/blog/PublishArticleForm.tsx
```diff
@@ -1,21 +1,32 @@
...
+  const { data: session } = useSession();
+  const isAdmin = session?.user?.role === UserRole.admin;
   const router = useRouter();
+  const tagManagerRef = useRef<TagManagerRef>(null);
+  const [selectedTags, setSelectedTags] = useState<Tag[]>([]);
+  const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
+
+  const addNewPostMutation = useAddNewPost(() => {
+    setHasUnsavedChanges(false);
+    router.push('/content-panel/blogs');
```
dichoskaemilija: here you can handle onSuccess and onError and return appropriate toast messages.
So you can follow the following pattern:
First define the mutation: const {mutate: addNewPost}= useAddNewPost()
So onSuccess => do this:
  const addNewPostMutation = (arguments) => {

    useAddNewPost(
      {
       arguments
      },
      {
        onSuccess: () => {
    setHasUnsavedChanges(false);
    router.push('/content-panel/blogs');
        },
        onError: (error: any) => {
    //show a toast 
        },
      },
    );
  };
  
 **This way the mutation hook will remain reusable and the side effects are localized**

components/module-components/blog/PublishArticleForm.tsx
```diff
@@ -31,119 +42,189 @@ const PublishArticleForm = () => {
...
+      }
+      return undefined;
+    };
+
+    window.addEventListener('beforeunload', handleBeforeUnload);
+    return () => window.removeEventListener('beforeunload', handleBeforeUnload);
+  }, [hasUnsavedChanges]);
+
   const handleAddPost = (values: NewPost) => {
+    setHasUnsavedChanges(false); // Reset immediately since we're about to save
```
dichoskaemilija: remove comment

dichoskaemilija (changes requested):

components/module-components/blog/TagInput.tsx
mitkapanarin: Removed useImperativeHandle by lifting searchTag state to TagManager for better state management

dichoskaemilija (approved):
great work @mitkapanarin 👏

PR #507, opened 2025-04-16
fix: now the pagination state correctly updates when a change occurs

gogostanoev:
- when a user selects a blog post and then goes back, they return to the same page they were on before selecting the post.
- when the page is reloaded, it preserves the current page and shows the same one as before the reload.


dichoskaemilija (changes requested):
Interesting approach.
Ideally this should work for all tables, not just the blogs table.
Assuming content-panel continues to grow in the future with more pages, if the user is in the blogs page, went to page 20 while browsing=> then visits ex: tags => goes back to the blogs page, she/he would expect to come to page 1 not 20.

